Issue #184, opened 2025-11-22
Add a makefile / justfile / taskfile?


etiennebacher:
- clippy + format
- update docs (make_docs.R + copy changelog)
- maybe test

PR #185, opened 2025-11-23
feat: Add rule for `browser()`

### LINKED ISSUE #174: Check for browser()
I'm interested in contributing a lint rule checking for `browser()` (which presumably has no place in a release). If you're open to this I'll work towards a PR (unless you'd like to implement it yourself).

jonocarroll:
Hopefully I've done this right. I didn't add any suggested fix because I don't know what one could safely do - I originally thought perhaps just comment out the call but it's common to call `browser()` as part of an `if` statement, so that won't work.

I'm happy to iterate on this if there's something else it should include.

Closes #174

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Features
 
 - Add support for `list2df` rule (#179).
-- Add support for `browser` rule (#174, @jonocarroll).
+- Add support for `browser` rule (#185, @jonocarroll).
 
 ### Fixes
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -5,7 +5,7 @@
 ### Features
 
 - Add support for `list2df` rule (#179).
-- Add support for `browser` rule (#174, @jonocarroll).
+- Add support for `browser` rule (#185, @jonocarroll).
 
 ### Fixes
 


jonocarroll:
FWIW, I get the same `clippy` issues on the main branch, and the remaining fails appear to be due to not being able to write back to the PR.

crates/jarl-core/src/lints/browser/browser.rs
```diff
@@ -0,0 +1,56 @@
...
+/// ## Example
+///
+/// ```r
+/// do_something <- function(abc = 1) {
+///    xyz <- abc + 1
+///    browser()
+///    xyz
+/// }
+///
+/// ```
```
etiennebacher:
```suggestion
/// do_something <- function(abc = 1) {
///    xyz <- abc + 1
///    browser()     # This should be removed.
///    xyz
/// }
/// ```
```

crates/jarl-core/src/lints/browser/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+        expect_no_lint("# remember to remove browser()", "browser", None);
+        expect_no_lint("function(browser = 'firefox')", "browser", None);
+        expect_no_lint("function(tool = browser)", "browser", None);
+        expect_no_lint("use a web browser to view the file", "browser", None);
+        expect_no_lint("hi", "browser", None);
+    }
+
+    #[test]
+    fn test_lint_browser() {
+        //use insta::assert_snapshot;
```
etiennebacher:
```suggestion
```

docs/rules.qmd
```diff
@@ -48,6 +48,7 @@ dat <- as.data.frame(
     c("any_duplicated", "Performance", "✅", ""),
     c("any_is_na", "Performance", "✅", ""),
     c("assignment", "Readability", "✅", ""),
+    c("browser", "Correctness", "✅", ""),
```
etiennebacher: This doesn't have a fix so the icon should be the red cross.

etiennebacher (changes requested):
Thanks @jonocarroll, it looks great!

Overall the implementation is good, but I have a few minor comments to fix below. Can you also add a bullet point in `CHANGELOG.md`? You'll also need to `cd docs && quarto render` (I will have more tooling to make it easier to apply those changes, cf #184).

> FWIW, I get the same clippy issues on the main branch

`cargo clippy` passes but `cargo fmt` fails, have you run it?

> the remaining fails appear to be due to not being able to write back to the PR

That's on me, there's probably something to fix in the CI workflows. I'll try to fix this before merging this PR because I want to see if there are any ecosystem changes.

etiennebacher:
> I didn't add any suggested fix because I don't know what one could safely do 

I don't think we should have an automated fix for this, it would be super annoying to have those statements removed if we have "Fix on Save" feature later on.

etiennebacher:
>> the remaining fails appear to be due to not being able to write back to the PR
>
> That's on me, there's probably something to fix in the CI workflows. I'll try to fix this before merging this PR because I want to see if there are any ecosystem changes.

I have added some changes in the workflows, please rebase or merge the main branch once you have addressed the comments above.

crates/jarl-core/src/lints/browser/browser.rs
```diff
@@ -0,0 +1,56 @@
...
+///
+/// Checks for lingering presence of `browser()` which should not be present in 
+/// released code. Does not remove the call as it does not have a suitable 
+/// replacement. One option would be `NULL` but this is possibly also bad.
+///
+/// ## Why is this bad?
+///
+/// `browser()` interrupts the execution of an expression and allows the inspection 
+/// of the environment where `browser()` was called from. This is helpful while 
+/// developing a function, but is not expected to be called by the user.
```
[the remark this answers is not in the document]
jonocarroll: ✅ Done

crates/jarl-core/src/lints/browser/browser.rs
jonocarroll: ✅ Done

crates/jarl-core/src/lints/browser/mod.rs
jonocarroll: ✅ Done

I did not add any snapshots, but I see that several other lints did so - this can easily be added back if those are necessary.

docs/rules.qmd
jonocarroll: ✅ Done

crates/jarl-core/src/lints/browser/browser.rs
```diff
@@ -0,0 +1,56 @@
...
+    }
+    fn body(&self) -> String {
+        "Calls to `browser()` should be removed.".to_string()
+    }
+}
+
+pub fn browser(ast: &RCall) -> anyhow::Result<Option<Diagnostic>> {
+    let RCallFields { function, .. } = ast.as_fields();
+
+    let function = function?;
```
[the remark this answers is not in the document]
etiennebacher: Thanks for the feedback, I'm currently writing https://github.com/etiennebacher/jarl/pull/183 so I'll take this into account. Would you be interested in reviewing it once I have completed the first draft? No problem if you don't want to.

crates/jarl-core/src/lints/browser/mod.rs
etiennebacher: No need here, snapshots are only necessary when there is an automated fix.

jonocarroll:
> > I didn't add any suggested fix because I don't know what one could safely do
> 
> I don't think we should have an automated fix for this, it would be super annoying to have those statements removed if we have "Fix on Save" feature later on.

I thought for a moment that there _were_ automated fixes, but I'm clearly confusing some other ecosystem. In any case, there's no good suggestion to make beyond "don't do that", IMHO.

I've hopefully updated appropriately. If there's more to fix I'll push more commits.

jonocarroll:
`cargo clippy` warns of various unrelated issues, identical to the main branch.

etiennebacher:
Perfect, thank you!

I have simply updated the PR title, I try to follow [Conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) messages. It allows to quickly distinguish commits in the tree, especially with the "Refined Github" extension.

PR #198, opened 2025-11-25
feat: Add `--no-default-exclude` cli argument

### LINKED ISSUE #178: running in a codebase with renv
So I am running this in a codebase with `renv` enabled:

```
jarl check . --output-format concise
renv\activate.R [221:7] implicit_assignment Avoid implicit assignments in function calls.
renv\activate.R [233:7] implicit_assignment Avoid implicit assignments in function calls.
renv\activate.R [1283:9] implicit_assignment Avoid implicit assignments in `if()` statements.
```

I  think I should not be getting any warnings about whatever it is that is happening with `renv` itself?

novica:
Use default-exclude patterns even without jarl.toml and add CLI arg --no-default-exclude

Adds a new argument in args.rs and check for the default-exclude files in check.rs. test checks that using the argument a file that should not be checked is checked and results in fixable suggestions.

Fixes #178


crates/jarl/tests/integration/no_default_exclude.rs
```diff
@@ -0,0 +1,32 @@
...
+fn test_no_default_exclude() -> anyhow::Result<()> {
+    let directory = TempDir::new()?;
+    let directory = directory.path();
+
+    let test_path = "cpp11.R";
+    let test_contents = "
+x = 1
+y <- 2
+3 -> z
+";
```
etiennebacher:
```suggestion
    let test_contents = "x = 1";
```

We just need one violation here.

crates/jarl/tests/integration/no_default_exclude.rs
```diff
@@ -0,0 +1,32 @@
...
+
+    insta::assert_snapshot!(
+        &mut Command::new(binary_path())
+            .current_dir(directory)
+            .arg("check")
+            .arg(".")
+            .arg("--no-default-exclude")
+            .run()
+            .normalize_os_executable_name()
+    );
```
etiennebacher: Can you add the same call without `no-default-exclude` first? We want to be sure that no violation is reported.

crates/jarl/tests/integration/snapshots/integration__no_default_exclude__no_default_exclude.snap
```diff
@@ -0,0 +1,21 @@
...
+  | --- Use `<-` for assignment.
+  |
+
+Found 1 error.
+1 fixable with the `--fix` option.
+
+----- stderr -----
+
+----- args -----
+check .
```
etiennebacher: It seems that this doesn't work as expected. If we exclude the list of patterns by default, then `cpp11.R` should be ignored by default and there shouldn't be any violation reported here.

etiennebacher (changes requested):
The default behaviour is wrong here because we never apply the list of patterns to exclude to our input paths. See the more detailed comment below.

crates/jarl/tests/integration/snapshots/integration__no_default_exclude__no_default_exclude.snap
novica: I am confused about something. `discover_r_file_paths()` is used in `jarl-lsp/src/lint.rs`. I obviously don't know how it is all tied together :) Does that have access to CLI arguments? 

Wouldn't it make sense that the paths are discovered from the `discover_settings`?

etiennebacher: > discover_r_file_paths() is used in jarl-lsp/src/lint.rs. I obviously don't know how it is all tied together :) Does that have access to CLI arguments?

This is true, but when we call `discover_r_file_paths()` in the LSP we will hardcode the value of the new argument so that violations in files like `cpp11.R` are not highlighted by default. If one wants to highlight those violations, then using `jarl.toml` will be the way to go.

So, in summary:

- in the CLI:
  * no `jarl.toml`:
    - by default, we exclude files like `cpp11.R`
    - if the user passes `--no-default-exclude`, we include files like `cpp11.R`
  * if the user has a `jarl.toml`:
    - if the TOML has `default-exclude = true`, then we exclude the files  
    - if the TOML has `default-exclude = true` and the CLI has `--no-default-exclude`, then include the files  
    - if the TOML has `default-exclude = false`, then we include the files, no matter whether  `--no-default-exclude` is passed

- in the LSP:
  * no `jarl.toml`: we always exclude files like `cpp11.R`
  * with `jarl.toml`:
    -  if it has `default-exclude = true`, we exclude these files
    -  if it has  `default-exclude = false`, we include these files (so they get highlighting)

You will need to update the call  to  `discover_r_file_paths()` in the LSP, but only pass a hardcoded `true` or `false` depending on the argument name. We can handle the rest of the LSP behavior in a separate PR.

> Wouldn't it make sense that the paths are discovered from the discover_settings?

I don't think so because in `discover_settings()` you would need to address the cases where a TOML exist and where a TOML doesn't exist. I think it's better to discover the (potentially non-existing) settings in the TOML first, and then resolve the paths based on those settings and on the value of `--no-default-exclude`.

novica: I think I am missing something obvious, or I am just getting confused by the opposite true/false options in the logic.

crates/jarl-core/src/discovery.rs
```diff
@@ -126,8 +127,13 @@ pub fn discover_r_file_paths<P: AsRef<Path>>(
...
 
-        // Check if default_exclude is disabled (true by default)
-        let use_default_exclude = settings.linter.default_exclude.unwrap_or(true);
+        // If the CLI requested "no default exclude", that always disables the DEFAULT_EXCLUDE_PATTERNS.
+        // Otherwise use the TOML value if present, or true if absent.
+        let use_default_exclude = if no_default_exclude {
+            false
+        } else {
+            settings.linter.default_exclude.unwrap_or(true)
+        };
```
etiennebacher: The problem here is that this chunk is located in this `if` condition:

```rust
if use_linter_settings && let Some(settings_item) = resolver.items().first() {
...
```

and if there's no `jarl.toml`, then `resolver.items().first()` is `None` and the code with `use_default_exclude` is never reached. I think this `if` condition should be modified to be something like this:

```rust
if use_linter_settings {
  // Build custom ignore patterns
  let mut patterns = Vec::new();

  if let Some(settings_item) = resolver.items().first() {
    // <existing code to handle settings>
   } else {
      if !no_default_exclude {
            // Add default exclude patterns
            patterns.extend_from_slice(DEFAULT_EXCLUDE_PATTERNS);
        }
      }
  }

  if !patterns.is_empty() { ... }
}
```

novica: OK. I think it works as it should now.

etiennebacher (changes requested):
Almost good indeed, there's just one small comment to address below, and you need to update `CHANGELOG.md`. After this is done, run `just document` to update `docs/changelog.md` as well.

Can you also mark this PR as ready to review?

It's also unexpected that the ecosystem check reports no changes. I checked manually that at least `dplyr` should have less fixes. I'll investigate that before merging.

novica:
> Almost good indeed, there's just one small comment to address below, and you need to update `CHANGELOG.md`.

I think this would be a breaking change, right? Because it changes the existing default behaviour.

etiennebacher:
Sorry for the extra commits, I can't figure out why the ecosystem checks still show no change but clearly this PR works locally, at least on `dplyr`.

Thanks @novica !

Issue #160, opened 2025-11-11
Enable "Fix on save" in editors


etiennebacher:
Ruff has a code action "Fix all" that can run on save: https://marketplace.visualstudio.com/items?itemName=charliermarsh.ruff

jrosell:
Thanks for this. I was about to ask how to do "Fix on save" and saw that it's already posted.

jrosell:
Do you think something like this could work?

    "[r]": {
        "editor.formatOnSave": true,
        "editor.defaultFormatter": ["etiennebacher.jarl-vscode", "Posit.air-vscode"]
    }

Reference: https://github.com/microsoft/vscode/issues/84603#issuecomment-553302611

etiennebacher:
Thanks for the suggestion. Just to be transparent, this is not a priority to me for now, and before thinking about the implementation it would be good to list how other linters approach this.

PR #215, opened 2025-11-28
docs: Neovim - Add section on enabling fix-on-save with Conform.nvim and remove typo

bjyberg:
I was able to get fix on save working in Neovim and thought it may be helpful to add in the editor documentation for people. It does require an additional Neovim plugin, but it is probably the most popular plugin for formatting in Neovim. I'll let you decide if it is worth adding! Somewhat of a cheat method of addressing #160 but it works well for me!

I also removed the note in parentheses about seeing the below note that no longer exists.

### DIFF
--- a/docs/editors.md
+++ b/docs/editors.md
@@ -125,8 +125,29 @@ vim.lsp.config('jarl', {})
 vim.lsp.enable 'jarl'
 ```
 
-This enables the code-actions and diagnostics (somewhat - see below note).
+This enables the code-actions and diagnostics.
 
 ![](img/nvim_diagnostic.png){fig-alt="R script with multiple errors showing in-line indicating a rule violation."}
 
 ![](img/nvim_quick_fix.png){fig-alt="The same R script as before, but this time there is a list of three actions next to the piece of code: apply fix, ignore this rule, and ignore all rules."}
+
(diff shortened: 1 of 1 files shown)


etiennebacher:
Hi @bjyberg, thanks for this but I'd like to have a better idea what the "Fix on save" behavior would be before starting documenting it. Currently, there are two points that bother me a bit:

- how would the "Fix on save" behavior of Jarl interact with other "<action> on save" provide by other tools, for instance "Format on save" from Air. Does the order of actions depend on the order of the tools if the config? Jarl is not a formatter so some fixes may lead to code that would be modified by a proper formatter like Air, so ideally Air must always run last.
- how would this interact with the absence of a VCS or with a dirty environment? I see that you mention `--allow-no-vcs` and we could also add `--allow-dirty` but this seems to be a global option that would be active for all projects, which is not desirable IMO.

I want to have this feature at some point, but maybe the way to go will be to implement the LSP feature "Fix all" and then have all other tools call this when saving the files (like this: https://jackevans.bearblog.dev/neovim-autofix-ruff-errors-on-save/). This is not my priority at the moment, but let's keep this open so it is somewhat visible to people who would be interested in this feature.

PR #252, opened 2025-12-29
[DO NOT MERGE] feat: Add rule `comparison_double`

### LINKED ISSUE #251: Add a rule to recommend using `all.equal()` instead of `==` on doubles?
There is this famous example:

``` r
x <- 0.1 * 3
x == 0.3
#> [1] FALSE
all.equal(x, 0.3)
#> [1] TRUE
```

and a lot of hits in CRAN packages: https://github.com/search?q=lang%3AR+%2F%3D%3D+%5Cd%5C.%5Cd%2B%2F+owner%3Acran&type=code

but I'm concerned about this making a lot of noise. Could be opt-in only?

Notes:

* AFAIK this is not present in `lintr`, I could suggest it there
* The fix should be `isTRUE(all.equal(x, value))`
* can only fix cases comparing to a single double value, not a vector (see SO link below)

Nice read: https://stackoverflow.com/a/9508558

etiennebacher:
Would fix #251 but want to see the potential impact first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 - New CLI argument `--statistics` to show the number of violations per rule instead
   of the details of each violation (#250).
 
+- New rule:
+  - `comparison_double` (#252)
+
 ## 0.3.0
 
 ### Breaking changes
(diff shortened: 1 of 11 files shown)


etiennebacher:
https://github.com/etiennebacher/jarl/issues/251#issuecomment-3696612065

PR #290, opened 2026-01-28
chore: Fix Alpine current stable release in integrations docs

philipp-baumann:
Hey @etiennebacher I noticed that I wrongly put `alpine:2.23` which should be `alpine:3.23` for a current stable release.


etiennebacher:
Thanks

PR #311, opened 2026-01-31
feat: Implement `if_always_true` rule

### LINKED ISSUE #237: Detect `if` conditions that are always true or false
All results are for code in the CRAN mirror on Github and include code that is commented out.

`if (TRUE)`: 428 hits

https://github.com/search?q=lang%3AR+%2Fif+%5C%28TRUE%5C%29%2F+org%3ACRAN+&type=code

`if (FALSE)`: 1.4k hits

https://github.com/search?q=lang%3AR+%2Fif+%5C%28FALSE%5C%29%2F+org%3ACRAN+&type=code

`if (TRUE || ...)`: 13 hits

https://github.com/search?q=lang%3AR+%2Fif+%5C%28TRUE+%5C%7C%5C%7C%2F+org%3ACRAN+&type=code

`if (FALSE && ...)`: 52 hits

https://github.com/search?q=lang%3AR+%2Fif+%5C%28FALSE+%26%26%2F+org%3ACRAN+&type=code

bjyberg:
I took a swing at implementing a linter to check where if statements are always true or false (#237).

Not Covered:
- the use of ifelse at all (i.e. ifelse(TRUE, ....) and so on)
- the edge case where if (TRUE > FALSE) or similar could be used

Feel free to let me know of any changes or suggestions. It does not add a quick-fix for anything

Fixes #237

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,11 +38,11 @@
 - New rules:
   - `equals_nan` (#284)
   - `equals_null` (#283)
+  - `if_always_true` (#311, @bjyberg)
   - `internal_function` (#291)
   - `redundant_ifelse` (#260)
   - `unnecessary_nesting` (#268)
   - `unreachable_code` (#261)
-  - `if_always_true` (#237)
 
 - When the output format is `full` or `concise`, rule names now have a hyperlink
   leading to the website documentation (#278).
--- a/crates/jarl-core/src/lints/base/if_always_true/if_always_true.rs
+++ b/crates/jarl-core/src/lints/base/if_always_true/if_always_true.rs
@@ -6,17 +6,17 @@ pub struct IfAlwaysTrue;
 
 /// ## What it does
 ///
-/// Detects `if` conditions that are always `TRUE`.
-/// This is only triggered for `if` statements without an `else`
-/// clause, these are handled by `unreachable_code`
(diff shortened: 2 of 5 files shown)


etiennebacher:
> perhaps this should just be an always_true linter and all of the logic for false can be handed in unreachable code? What do you think?

Yes, that sounds good to me.

> And nice documentation for this by the way, it was very helpful!

Happy to read that! (I'll have to slightly tweak the guide following the recent refactor)

etiennebacher:
@bjyberg just a quick heads-up, I will probably release 0.4.0 during at the end of the incoming week (still need to finish a couple of things and draft a blog post). No pressure at all from my side, just letting you know in case you want this PR in the next release.

etiennebacher:
> would you prefer I strip out all code related to checking if a value is false?

Yes you can remove it, I'd rather not add code that is not relevant to the rule.

bjyberg:
Nice great, thanks for the help @etiennebacher! Haha yes, it ended up being a bit tougher than I expected, but good learning experience for me! Keep me in the loop if you can think of any other rules with a similar or easier difficulty level, and I'll try to find some time to work on a few.

Good to know regarding the docs/md files.. I probably should have assumed that. I'll remember that in the future. Cheers!

Issue #214, opened 2025-11-27
Potential additional rules for `testthat`


etiennebacher:
Example PR: https://github.com/etiennebacher/jarl/pull/204
Detailed contributing guide: https://jarl.etiennebacher.com/contributing-rule-example

For those that end up being implemented, also suggest them to `lintr` (with a link to the Jarl tests):

- [x] [`expect_match()`](https://testthat.r-lib.org/dev/reference/expect_match.html) instead of `expect_true(grepl(...))`
- [x] [`expect_no_match()`](https://testthat.r-lib.org/dev/reference/expect_match.html) instead of `expect_false(grepl(...))`
- [ ] [`expect_in()`](https://testthat.r-lib.org/dev/reference/expect_setequal.html) instead of `expect_true(all(object %in% expected))`
- [ ] [`expect_contains()`](https://testthat.r-lib.org/dev/reference/expect_setequal.html) instead of `expect_true(all(expected %in% object))`

PR #364, opened 2026-02-23
feat: Add expect_match linter (testthat)

bjyberg:
Addresses part of #214.

Add a linter + fix for  `expect_match()` instead of `expect_true(grepl(...))`

It handles pipes following the recent update, and I think I got the tests consistent with the recent updates as well.

I assume that the `expect_no_match` linter will be incredibly similar to this, so I can probably get that one done quite soon after.

## Fix
This does not apply a fix for pipes as the code started getting fairly complex when I was attempting it due to the positions of x and pattern being different between `grepl` and `expect_match`. It also skips the fix when additional args are passed to `expect_true` (info and msg) - these seem to be able used in expect_match as well, but the other linters in testthat seem to skip these, so that is the route I went. 

The fix is used for additional args passed to grepl, as these can all be passed to expect_match.

### DIFF
--- a/crates/jarl-core/src/lints/testthat/expect_match/expect_match.rs
+++ b/crates/jarl-core/src/lints/testthat/expect_match/expect_match.rs
@@ -1,7 +1,7 @@
 use crate::diagnostic::*;
 use crate::utils::{
-    drop_arg_by_name_or_position, get_arg_by_name_then_position, get_function_name,
-    get_function_namespace_prefix, get_nested_functions_content, node_contains_comments,
+    get_arg_by_name_then_position, get_function_name, get_function_namespace_prefix,
+    get_nested_functions_content, node_contains_comments,
 };
 use air_r_syntax::*;
 use biome_rowan::{AstNode, AstSeparatedList};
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -55,6 +55,7 @@
   - `duplicated_function_definition` (#358)
   - `invalid_chunk_suppression` (#350)
   - `undesirable_function` (replaces `browser`) (#336)
+  - `expect_match` (#214, @bjyberg)
```
etiennebacher: Please put it in alphabetical order

crates/jarl-core/src/lints/testthat/expect_match/mod.rs
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    #[test]
+    fn test_expect_match_with_comments_no_fix() {
+        assert_snapshot!(
+            "no_fix_with_comments",
+            get_fixed_text(
+                vec![
+                    "expect_true(grepl(# comment\n'fun', 'Testing is fun'))",
+                    "expect_true(grepl('fun', 'Testing is fun', # comment\nperl = TRUE))",
```
etiennebacher:
```suggestion
```

I don't think this line tests something new compared to the one above.

PR #381, opened 2026-03-06
feat: Add `quotes` rule

bjyberg:
I may have been a bit ambitious with my abilities on this one, but I think I got there in the end. I often find myself missing the quote style lint and fix from Lintr and Styler, so that gave me some motivation to try and implement it.

It should implement the same behavior as the [Quotes linter from lintr](https://lintr.r-lib.org/reference/quotes_linter.html). I did a lot of copying from the assignment linter, undesirable function linter, etc. I took most of the tests from Linter as well.

Let me know what you think, any changes, whether you want to include this now or not, etc.

### DIFF
--- a/crates/jarl-core/src/lints/base/quotes/mod.rs
+++ b/crates/jarl-core/src/lints/base/quotes/mod.rs
@@ -49,7 +49,7 @@ mod tests {
          --> <test>:1:1
           |
         1 | 'hi'
-          | ---- Prefer double-quotes for string delimiters.
+          | ---- Prefer double quotes for string delimiters.
           |
         Found 1 error.
         "
@@ -61,7 +61,7 @@ mod tests {
          --> <test>:1:5
           |
         1 | fun('hello')
-          |     ------- Prefer double-quotes for string delimiters.
+          |     ------- Prefer double quotes for string delimiters.
           |
         Found 1 error.
         "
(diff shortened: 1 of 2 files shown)


crates/jarl-core/src/toml.rs
```diff
@@ -220,6 +221,13 @@ pub struct LinterTomlOptions {
     #[serde(rename = "implicit_assignment")]
     pub implicit_assignment: Option<ImplicitAssignmentOptions>,
 
+    /// # Options for the `quotes` rule
+    ///
+    /// Use `quote` to choose the preferred quote delimiter for string
+    /// literals. Valid values are `"double"` (default) and `"single"`.
+    #[serde(rename = "quotes")]
+    pub quotes: Option<QuotesOptions>,
```
etiennebacher: Should be named `quote` I think? If not, the comment above needs to be updated

etiennebacher (commented):
Looks really great, nice one! I kinda wish this was in Air so it can fix my quotes when I save the file, but it's already nice to have it here.

---

A few more comments:

- in this example, the message "use double-quotes only" is ambiguous because it could also be related to the `\'` (but it's not):

  ```
   --> test.R:3:1
    |
  3 | '\'a'
    | ----- Only use double-quotes.
    |
  ```
  Can we improve the message a bit? Like `Use double quotes to start and end a string`? I'm really not sure this is better, so open to alternatives. 

* maybe it's worth reporting unnecessarily escaped quote, such as `"\'a"` that could simply be `"'a"`. I don't know if there's a risk of false positives, we'd need to see ecosystem results for that. This doesn't have to be done in this PR, it's just a suggestion for an extension.

PR #458, opened 2026-04-22
docs: YAML formatting for CI branch specification

JosephBARBIERDARNAL:
Hi! 

Just a small PR to update CI example in the doc. `branches` expects an array rather than a single element, which triggers a message from some GA linters, and might break one day (unlikely but better be safe than sorry):

image


etiennebacher:
Nice catch, thanks!

Issue #239, opened 2025-12-09
Recommend `%notin%` operator


etiennebacher:
After R 4.6.0 is out.

Yousa-Mirage:
Maybe I can try to complete this rule?

etiennebacher:
Sure, feel free to give it a try. Just keep the changes for `equals_na` and `any_is_na` in separate PRs.

Yousa-Mirage:
OK. I might ask you some questions about the rule detail and implementation here.

etiennebacher:
Sure, no problem 👍 

The tutorial should help here: https://jarl.etiennebacher.com/contributing-rule-example

Yousa-Mirage:
Do you think we need and have a way to deal with redundant multiple layers of parentheses, such as'! ((((x %in% y))))`?

etiennebacher:
There could be a new rule to find unnecessary parenthesis (for reference, this is reported by Ruff with code UP034), so that's not related to the `%notin%` rule. I think it's just a matter of checking if the first and only child of a `R_PARENTHESIZED_EXPRESSION` is also a `R_PARENTHESIZED_EXPRESSION`.

PR #459, opened 2026-04-25
feat: add `notin` rule

Yousa-Mirage:
Part of #239


crates/jarl-core/src/rule_set.rs
```diff
@@ -440,6 +440,13 @@ declare_rules! {
         fix: Safe,
         min_r_version: None,
     },
+    Notin => {
```
etiennebacher: This should be `NotIn`

crates/jarl-core/src/lints/base/notin/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+    use insta::assert_snapshot;
+
+    fn snapshot_lint(code: &str) -> String {
+        format_diagnostics(code, "notin", Some("4.6"))
+    }
+
+    #[test]
+    fn test_lint_notin() {
+        assert_snapshot!(
+            snapshot_lint("!(x %in% y)"),
```
etiennebacher, line 15: Can you add a test for `!x %in% y`?

crates/jarl-core/src/lints/base/notin/notin.rs
```diff
@@ -0,0 +1,95 @@
...
+/// ## References
+///
+/// See `?match`
+impl Violation for Notin {
+    fn name(&self) -> String {
+        "notin".to_string()
+    }
+    fn body(&self) -> String {
+        "`!(x %in% y)` can be simplified.".to_string()
+    }
```
etiennebacher: Based on the comment above, this comment should be adapted to not show parenthesis if there are none in the user's code. The comment to report to the user must be created  in the `notin()` function below.

crates/jarl-core/src/lints/base/notin/notin.rs
```diff
@@ -0,0 +1,95 @@
...
+use air_r_syntax::*;
+use biome_rowan::AstNode;
+
+pub struct Notin;
+
+/// Version added: 0.6.0
+///
+/// ## What it does
+///
+/// Checks for usage of `!(x %in% y)` and recommends using `%notin%` instead.
```
etiennebacher:
```suggestion
/// Checks for usage of `!x %in% y` and recommends using `%notin%` instead.
```

docs/changelog.md
```diff
@@ -2,6 +2,11 @@
 
 ## Development version
 
+### Features
+
+* New rule `notin` recommends `%notin%` instead of `!(x %in% y)` for projects
+  using R >= 4.6.0 (#459).
```
etiennebacher: Just list the new rule, as in the changelog of previous versions (you can also add your github username if you want).

docs/rules.qmd
```diff
@@ -95,6 +95,7 @@ dat <- as.data.frame(
     c("misplaced_file_suppression", "comments", "❌", ""),
     c("misplaced_suppression", "comments", "❌", ""),
     c("numeric_leading_zero", "readability", "✅", ""),
+    c("notin", "readability", "✅", "R >= 4.6"),
```
etiennebacher, line 97: Please follow alphabetic order

etiennebacher (changes requested):
Looks good, thanks! A few minor comments below.

Yousa-Mirage:
OK. It seems that I don't have a good grasp of the order of R operators myself, I thought `!x %in% y == (!x) %in% y` instead of `!(x %in% y)`, so I only checked the case with parentheses.

etiennebacher:
FYI, you can check `?Syntax` to get the full precedence order in R.

Yousa-Mirage:
In the official documention in `?Syntax`, they say:
```r
## Special operators have higher precedence than "!" (logical NOT).
## You can use this for %in% :
! 1:10 %in% c(2, 3, 5, 7) # same as !(1:10 %in% c(2, 3, 5, 7))
## but we strongly advise to use the "!( ... )" form in this case!
```
Looks like I'm not the only one ignorant :)

**Do we need to add a rule that prompts for adding parentheses in expressions like `! 1:10 %in% c(2, 3, 5, 7)`?**

etiennebacher:
> Do we need to add a rule that prompts for adding parentheses in expressions like ! 1:10 %in% c(2, 3, 5, 7)?

No, I don't think so, it's quite common not to use parenthesis here and now `%notin%` fills this gap IMO.

etiennebacher (approved):
Thanks!

PR #461, opened 2026-04-26
fix: wrong fix when having internal comments

### LINKED ISSUE #460: Bug: leading & trailing comments will result in the loss of internal comments
Like this file:
```r
# leading comment
!(x
  # VERY IMPORTANT COMMENT!!!
  >= y)

!(x
  # VERY IMPORTANT COMMENT!!!
  >= y) # trailing comment
```

After run `jarl check --fix`, it will become this:
```r
# leading comment
x < y

x < y # trailing comment
```

The comments in the middle are lost.

This is due to this incorrect boolean logic in `crates/jarl-core/src/utils.rs`:
```rust
pub fn node_contains_comments(node: &air_r_syntax::RSyntaxNode) -> bool {
    (node.has_comments_direct() || node.has_comments_descendants())
        && !node.has_trailing_comments()
        && !node.has_leading_comments()
}
```
The original intention of this function is to ignore `trailing_comments` and `leading_comments` and only check the internal comments. However, due to a Boolean logic error, whenever `trailing_comments` or `leading_comments` appear, the function will always return `false`, and thus cannot skip the fix.

I will open a PR to fix this issue.

Yousa-Mirage:
fix #460 

When auto-fix, this code will lost the internal comment:
```r
# leading comment
!(x
  # VERY IMPORTANT COMMENT
  >= y)

!(x
  # VERY IMPORTANT COMMENT
  >= y) # trailing comment
```

Fix result before (lost comment):
```r
# leading comment
x < y

x < y # trailing comment
```

Fix result after (no change):
```r
# leading comment
!(x
  # VERY IMPORTANT COMMENT
  >= y)

!(x
  # VERY IMPORTANT COMMENT
  >= y) # trailing comment
```

I changed `node_contains_comments()` to token level judgment. The current logic can be done correctly:
  - Fix is not skipped when there are only pre and post comments
  - Skip fix as long as there is a comment inside
I also added unit tests for `node_contains_comments()` and tests for `comparison_negation` rules.

All tests passed.


crates/jarl-core/src/utils.rs
```diff
@@ -393,8 +393,59 @@ pub fn get_nested_functions_content(
...
+    node.descendants_tokens(Direction::Next).any(|token| {
+        let has_internal_leading =
+            token.has_leading_comments() && first_token.as_ref() != Some(&token);
+        let has_internal_trailing =
+            token.has_trailing_comments() && last_token.as_ref() != Some(&token);
+        has_internal_leading || has_internal_trailing
+    })
+}
+
+#[cfg(test)]
```
etiennebacher: Usually I tend to prefer testing the external behavior of the tool rather than internal functions, so you can remove this block.

docs/changelog.md
```diff
@@ -4,6 +4,9 @@
 
 ### Bug fixes
 
+* Automatic fixes now skip rewrites when comments are inside the rewritten
+  expression, while still allowing leading and trailing comments (#461).
```
etiennebacher: You can also add your Github username if you want

crates/jarl-core/src/lints/base/comparison_negation/mod.rs
```diff
@@ -175,6 +175,7 @@ mod tests {
                 vec![
                     "# leading comment\n!(x >= y)",
                     "!(x \n # hello there \n >= y)",
+                    "# leading comment\n!(x \n # hello there \n >= y)",
```
etiennebacher: This is a general bug fix, so the test for this shouldn't be in the tests of a specific rule. Please add a test in `crates/jarl/tests/integration/comments.rs` instead.

etiennebacher (changes requested):
Thanks for catching the bug and creating a fix, this looks good to me. Just a couple of comments related to tests.

crates/jarl/tests/integration/comments.rs
```diff
@@ -332,3 +332,46 @@ any(is.na(z))
 
     Ok(())
 }
+
+#[test]
+fn test_fix_skips_internal_comments_with_outer_comments() -> anyhow::Result<()> {
```
etiennebacher:
```suggestion
fn test_fix_skips_internal_comments_with_outer_comments_460() -> anyhow::Result<()> {
```

PR #469, opened 2026-04-30
feat: Add check for `x %notin% NA` in `equals_na` rule

Yousa-Mirage:
Part of #239
Add check for `x %notin% NA` in `equals_na` rule.

### DIFF
--- a/crates/jarl-core/src/lints/base/equals_na/equals_na.rs
+++ b/crates/jarl-core/src/lints/base/equals_na/equals_na.rs
@@ -9,8 +9,8 @@ pub struct EqualsNa;
 ///
 /// ## What it does
 ///
-/// Check for `x == NA`, `x != NA` and `x %in% NA`, and replaces those by
-/// `is.na()` calls.
+/// Check for `x == NA`, `x != NA`, `x %in% NA` and `x %notin% NA`, and
+/// replaces those by `is.na()` or `!is.na()` calls.
 ///
 /// ## Why is this bad?
 ///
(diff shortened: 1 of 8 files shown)


crates/jarl-core/src/lints/base/equals_na/equals_na.rs
```diff
@@ -39,10 +39,10 @@ impl Violation for EqualsNa {
         "equals_na".to_string()
     }
     fn body(&self) -> String {
-        "Comparing to NA with `==`, `!=` or `%in%` is problematic.".to_string()
+        "Comparing to NA with `==`, `!=`, `%in%` or `%notin%` is problematic.".to_string()
```
etiennebacher: I feel like the body and suggestion could be defined in the function so that the message isn't full of operators that make it hard to read. This was already the case before this PR so it's not mandatory, but do you want to tackle that?

docs/changelog.md
```diff
@@ -7,6 +7,7 @@
 * New rules:
 
   - `notin` (#459, @Yousa-Mirage)
+  - `equals_na` now also reports `x %notin% NA` cases (#469, @Yousa-Mirage)
```
etiennebacher: Please put this in alphabetical order

etiennebacher (approved):
Thanks, LGTM overall, just a suggestion for a small extension.

etiennebacher (approved):
Perfect, thank you for doing the extra suggestion!

Issue #472, opened 2026-05-06
New rule: no empty files


JosephBARBIERDARNAL:
Hi!

I figure it would make sense to add a rule that checks for empty R files? I had one in a project and I wasn't aware of it, mostly because the number of files is large.

etiennebacher:
I can't find any downsides to it, but should files containing only comments be reported as well? Also, do you want to give it a try?

JosephBARBIERDARNAL:
Good question, I'm unsure. My intuition is yes, but no strong opinion.

And sure, I'll open a draft PR soon!

PR #477, opened 2026-05-11
feat: Add rule for empty files

JosephBARBIERDARNAL:
Draft PR for #472 

I'm still trying to understand the codebase, but I have some questions that need to be answered before I actually implement this:

* Are comments ignored? I changed my mind, and I think that if there are some comments, then this shouldn't be considered empty. Maybe this could be an option in the config file? But I doubt people would change this. This would also change whether the rule needs `RExpressionList` or not.
* I assume the rule should be checked in `check.rs::get_checks`?


JosephBARBIERDARNAL:
@etiennebacher I allow myself to ping you here in case you didn't see [my first message above](https://github.com/etiennebacher/jarl/pull/477#issue-4422885233)

etiennebacher:
Thanks, I had seen the original PR but not the edits you made after (I don't get a notification for edits). I'll try to take a look today or tomorrow

etiennebacher:
> Are comments ignored? I changed my mind, and I think that if there are some comments, then this shouldn't be considered empty. Maybe this could be an option in the config file? But I doubt people would change this. This would also change whether the rule needs RExpressionList or not.

It's possible someone comments out all the code in a file and forgets to uncomment it. IMO this  is more likely than someone  creating an R file containing only comments on purpose, so I would rather flag a file that only has comments. I'd rather avoid adding config since the case with only comments on purpose sounds extremely rare to me.

> I assume the rule should be checked in check.rs::get_checks?

I think you should rather add an `if` condition in `check_document()`. We create a vector of expressions at the top of this function and you can directly check if this vector is empty and if `checker.is_rule_enabled(Rule::EmptyFile)` is true.

You will also need to create a function in `jarl-core/src/lints/base` but this function will be extremely simple because it will just return a diagnostic: we know that if we reached this function then the file is empty so there's no more check to do.

I hope this helps and feel free to ping me if you're stuck somewhere ;)

JosephBARBIERDARNAL:
Great, that sounds good thanks!

crates/jarl-core/src/analyze/document.rs
```diff
@@ -129,6 +130,10 @@ pub(crate) fn check_document(
         }
     }
 
+    if checker.is_rule_enabled(Rule::EmptyFile) && expressions.is_empty() {
+        checker.report_diagnostic(Some(empty_file()));
+    }
+
```
JosephBARBIERDARNAL, line 137: AI suggested me to do this, but I feel like it would be more intuitive to pass `expressions` to `empty_file()` and check if it's empty inside the function? This would also make things a bit more consistent with how other rules are checked.

crates/jarl-core/src/lints/base/empty_file/empty_file.rs
```diff
@@ -0,0 +1,20 @@
...
+
+impl Violation for EmptyFile {
+    fn name(&self) -> String {
+        "empty_file".to_string()
+    }
+    fn body(&self) -> String {
+        "This file is empty or only contains comments.".to_string()
+    }
+    fn suggestion(&self) -> Option<String> {
+        Some("Consider deleting the file".to_string())
```
JosephBARBIERDARNAL: This message might be improved, because when one has a completly empty file, it only display that we should delete the file, not why:

```R
warning: [empty_file](https://jarl.etiennebacher.com/rules/empty_file)
--> test_empty.R:1:1
 |
 |
 = help: Consider deleting the file


Found 1 error.
```

crates/jarl-core/src/lints/base/empty_file/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+        // Comments alongside real code don't count as empty
+        expect_no_lint("# header\nx <- 1", "empty_file", None);
+
+        // Even a bare literal is content
+        expect_no_lint("NULL", "empty_file", None);
+
+        // Whitespace around a single expression is fine
+        expect_no_lint("\n\n  x <- 1  \n\n", "empty_file", None);
+    }
+}
```
JosephBARBIERDARNAL, line 102: Tests cover:
- completly empty file
- file with white space
- file with comments

They all raise a warning.

docs/rules/empty_file.md
```diff
@@ -0,0 +1,20 @@
...
+
+An empty or comment-only file is almost always a mistake: a placeholder that was forgotten, an accidental `touch`, or leftover from a refactor. It adds noise to the package and can confuse readers.
+
+## Example
+
+```r
+# TODO: implement the data loader
+```
+
+Use instead: delete the file, or add the intended code.
```
JosephBARBIERDARNAL: This is mostly AI generated, but looks fine to me.

docs/rules/empty_file.md
```diff
@@ -0,0 +1,20 @@
+# empty_file
+
+::: {.callout-note title="Added in 0.5.\*" .low-opacity}
```
JosephBARBIERDARNAL: I'm unsure of what will be the first version that contains the rule so I put this in the meantime

crates/jarl-lsp/src/lint.rs
```diff
@@ -413,7 +413,11 @@ mod tests {
     fn test_empty_document() {
         let snapshot = create_test_snapshot("");
         let output = lint_document(&snapshot).unwrap();
-        assert!(output.diagnostics.is_empty());
+        assert_eq!(output.diagnostics.len(), 1);
+        assert_eq!(
+            output.diagnostics[0].message,
+            "This file is empty or only contains comments. Consider deleting the file"
+        );
```
JosephBARBIERDARNAL, line 424: Now we expect to raise one diagnostic (e.g., empty file)

JosephBARBIERDARNAL:
@etiennebacher This is ready for review! I added a bunch of comments that hopefully make things clearer.

EDIT: 
- CI fails on Windows for the `test_empty_document()`, I'm unsure why (maybe some cache/state issue?)
- CI fails for `links`, also unsure what that is

crates/jarl-core/src/analyze/document.rs
etiennebacher: It makes sense to me to move this emptiness check to the function 👍

docs/rules/empty_file.md
etiennebacher: This docs should be written in the .rs file that contains the function. The md file is then generated with `just document`. See more details here: https://jarl.etiennebacher.com/contributing-rule-example

docs/rules/empty_file.md
etiennebacher: You can put 0.6.0, it will be the next version

etiennebacher:
Thanks! Don't worry about the CI for now, there are some issues coming from upstream.

I replied to a few comments but I'm on my phone so I'll make a proper review once I have access to a laptop.

crates/jarl-core/src/lints/base/empty_file/empty_file.rs
```diff
@@ -0,0 +1,49 @@
...
+/// ## What it does
+///
+/// Reports R files that contain no code: either truly empty, only whitespace,
+/// or only comments.
+///
+/// ## Why is this bad?
+///
+/// An empty or comment-only file is almost always a mistake: a placeholder that
+/// was forgotten, an accidental `touch`, or a leftover from a refactor. It adds
+/// noise to the package and can confuse readers.
```
etiennebacher:
```suggestion
/// An empty or comment-only file is almost always a mistake: a placeholder that
/// was forgotten, or a leftover from a refactor. It adds noise to the package and 
/// can confuse readers.
```

You will need to run again `just document`.

crates/jarl-core/src/lints/base/empty_file/empty_file.rs
```diff
@@ -0,0 +1,49 @@
...
+///
+/// ## Example
+///
+/// ```r
+/// # TODO: implement the data loader
+/// ```
+///
+/// Use instead: delete the file, or add the intended code.
```
etiennebacher:
```suggestion
/// Instead, delete the file or add the intended code.
```

crates/jarl-core/src/lints/base/empty_file/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+          |
+        1 | # just a comment
+          | - This file is empty or only contains comments.
+          |
+          = help: Consider deleting the file
+        Found 1 error.
+        ");
+
+        // Case 4: roxygen-style comments only
+        assert_snapshot!(snapshot_lint("#' @keywords internal"), @"
```
etiennebacher: After testing locally and given the results of the ecosystem, I think we should allow files with roxygen comments only since they can be used as templates (R files in `man-roxygen` for instance). 

This adds a little difficulty here because when we process the expression list in `check_document()` we already lost the comments in the file. Those can be recovered with `&syntax.text_with_trivia()` in `get_checks()`, so I suppose we should modify `check_document()` to to have an extra arg `syntax`, and we use this extra arg only in `empty_file()` (where we check if any comment starts with `#'`).

On top of that:

- this exception should also be documented
- there should be some tests with roxygen comments only, and with a mix of roxygen and non-roxygen comments

docs/changelog.md
```diff
@@ -10,6 +10,7 @@
   * `equals_na` now also reports `x %notin% NA` cases (#469, @Yousa-Mirage)
   * `notin` (#459, @Yousa-Mirage)
   * `pipe_consistency` (#482)
+  * `empty_file` (#477, @JosephBARBIERDARNAL)
```
etiennebacher: Please put this in alphabetical order

test_empty.R
etiennebacher, line 1: This was probably pushed by mistake

etiennebacher (changes requested):
Getting closer, but I think we need to skip this rule on files with roxygen comments only, see my comment below.

etiennebacher:
There should also be some tests to ensure that a `# jarl-ignore` comment in an otherwise empty file properly ignores this file, e.g.:

```
# jarl-ignore empty_file: this is on purpose
```

and 

```
# jarl-ignore empty_file: this is on purpose

# hello
# there
```

shouldn't be reported (this is already the case).

JosephBARBIERDARNAL:
Just updated with your requested changes. I'll implement the roxygen update later this week

crates/jarl-core/src/analyze/document.rs
```diff
@@ -18,6 +19,7 @@ use crate::rule_set::Rule;
 
 pub(crate) fn check_document(
     expressions: &RExpressionList,
+    syntax: &RSyntaxNode,
```
JosephBARBIERDARNAL, line 22: We need this to check for Roxygen comments

JosephBARBIERDARNAL:
Ready for another pass of review!

crates/jarl-core/src/lints/base/empty_file/empty_file.rs
```diff
@@ -0,0 +1,60 @@
...
+/// Instead, delete the file or add the intended code.
+impl Violation for EmptyFile {
+    fn name(&self) -> String {
+        "empty_file".to_string()
+    }
+    fn body(&self) -> String {
+        "This file is empty or only contains comments.".to_string()
+    }
+    fn suggestion(&self) -> Option<String> {
+        Some("Consider deleting the file".to_string())
```
etiennebacher:
```suggestion
        Some("Consider deleting the file.".to_string())
```

crates/jarl-core/src/lints/base/empty_file/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+        );
+
+        // Roxygen-only file (e.g. man-roxygen/ template) is allowed
+        expect_no_lint("#' @keywords internal", "empty_file", None);
+
+        // Multi-line roxygen-only file is allowed
+        expect_no_lint("#' some doc\n#' more docs", "empty_file", None);
+
+        // Mixed roxygen + plain comments is allowed (any `#'` exempts the file)
+        expect_no_lint("#' some doc\n# plain comment", "empty_file", None);
```
etiennebacher, line 94: Can you just add the same test but with the `#` coming first and then the `#'`?

etiennebacher (commented):
LGTM, just a couple of tiny comments to address. The test in `jarl-lsp` seems to constantly fail on windows so it is probably platform specific. I don't really know why but some AI tool might help here.

---

I'm wondering if we could have an escape hatch for cases where the person really wants an empty file (e.g. the results in `box` package) without entirely disabling this rule. We could add a configuration option `skipped-files` in the TOML. This adds some complexity so I can handle that if you prefer, just let me know.

crates/jarl-core/src/lints/base/empty_file/mod.rs
JosephBARBIERDARNAL: In that case, I guess jarl is supposed to trigger the rule right?

JosephBARBIERDARNAL:
> LGTM, just a couple of tiny comments to address. The test in `jarl-lsp` seems to constantly fail on windows so it is probably platform specific. I don't really know why but some AI tool might help here.

I'm working on a Mac, but I have a Windows machine I can use for tests, I'll see if I'm able to reproduce the issue.

JosephBARBIERDARNAL:
> I'm wondering if we could have an escape hatch for cases where the person really wants an empty file (e.g. the results in `box` package) without entirely disabling this rule. We could add a configuration option `skipped-files` in the TOML. This adds some complexity so I can handle that if you prefer, just let me know.

Isn't it what `exclude` already do?

crates/jarl-core/src/lints/base/empty_file/mod.rs
etiennebacher: No I don't think so because you might get a standard comment at the top of a template file.

etiennebacher:
> > I'm wondering if we could have an escape hatch for cases where the person really wants an empty file (e.g. the results in `box` package) without entirely disabling this rule. We could add a configuration option `skipped-files` in the TOML. This adds some complexity so I can handle that if you prefer, just let me know.
> 
> Isn't it what `exclude` already do?

Hmm, yes fair enough. I was thinking of [rule-specific arguments](https://jarl.etiennebacher.com/reference/config-file#rule-specific-arguments) but excluding the file entirely works since it's supposed to be empty. Problem solved then :)

---

I think you need to update the snapshots since the message got updated in the last commit.

JosephBARBIERDARNAL:
> > LGTM, just a couple of tiny comments to address. The test in `jarl-lsp` seems to constantly fail on windows so it is probably platform specific. I don't really know why but some AI tool might help here.
> 
> I'm working on a Mac, but I have a Windows machine I can use for tests, I'll see if I'm able to reproduce the issue.

Turns out it's a pain to make cargo work properly on Windows 😅 I'll see what AI thinks instead

crates/jarl-lsp/src/lint.rs
```diff
@@ -396,8 +396,8 @@ mod tests {
     use crate::session::DocumentSnapshot;
     use lsp_types::{ClientCapabilities, Url};
 
-    fn create_test_snapshot(content: &str) -> DocumentSnapshot {
-        let uri = Url::parse("file:///test.R").unwrap();
+    fn create_test_snapshot(file_path: &std::path::Path, content: &str) -> DocumentSnapshot {
+        let uri = Url::from_file_path(file_path).unwrap();
```
JosephBARBIERDARNAL, line 400: Both codex and claude code suggested this change, so that sounds fair to me. Let's see how the CI goes.

JosephBARBIERDARNAL: Update: looks like that was the issue.

etiennebacher:
Great, thanks for addressing all my comments!

PR #478, opened 2026-05-11
feat: Use Oak to get more semantic information

etiennebacher:
To compare to #451. It has the same objective but uses Oak's (young) interface developed in Ark.

Still very experimental, mostly here to get ecosystem checks and compare to the `flowr` approach. At a first glance it looks simpler to rely on the interface developed by Posit rather than translating the TS codebase of `flowr` into Rust.


etiennebacher:
Two failures related to `glue`:

1. Run those separately:

```r
# no violation
foo <- "a"
glue("{foo}")

# no violation
x <- "a"
glue("|x|", .open = "|", .close = "|")

# violation (shouldn't be)
foo <- "a"
glue("|foo|", .open = "|", .close = "|")
```

2. Run those together:

```r
# no violation
foo <- "a"
glue("{foo}")

# no violation (should be)
foo <- "a"
```

Also need check those for `cli`.

etiennebacher:
Run these separately:

```r
# no violation
`%op%` <- function() {}
1 %op% 2

# violation (shouldn't be)
f <- function() {}
`%op%` <- f
1 %op% 2
```

(`%op%` is surrounded by backticks in code)

PR #484, opened 2026-05-17
feat: Add new rule `glue`

### LINKED ISSUE #331: Find unnecessary `glue()` calls
e.g. `glue("my message")`

Many cases: https://github.com/search?q=lang%3AR+%2Fglue%5C%28%22%5B%5E%5C%7B%5D%2B%22%5C%29%2F+NOT+is%3Afork&type=code

This should be similar to some cases of the ["sprintf" rule](https://jarl.etiennebacher.com/rules/sprintf), but we need to handle the arguments `.open`, `.close`, and `.sep`: https://glue.tidyverse.org/reference/glue.html

`glue()` can also takes several pieces of text, e.g. `glue("a", "b")`.

novica:
This introduces a rule for glue. 

Copilot provided code and context on R string literals, and I am not sure how the rule should apply partly on R 4.0+ ?

Fixes #331


etiennebacher:
> I am not sure how the rule should apply partly on R 4.0+ ?

`glue` depends on R >= 4.1 anyway, but I prefer adding R version constraints for rules that depend on a language feature. Here, we'd need to update the version constraint every time it is updated in `glue`, so let's not put a constraint.

I just rebased this PR to clarify the ecosystem results.

novica:
Should I be doing something more here?

etiennebacher:
Apparently some tests are failing because snapshots need to be updated. Some ecosystem results also look incorrect to me (such as the first one of arrow). Can you double check?

novica:
I think it looks ok now? But honestly have no idea what the codecov report is trying to say :)

crates/jarl-core/src/lints/base/glue/glue.rs
```diff
@@ -0,0 +1,207 @@
...
+///
+/// ## What it does
+///
+/// Multiple checks for `glue()`:
+///
+/// 1. checks whether `glue()` has exactly one unnamed argument and no other named
+///    arguments (in which case `glue()` is not needed);
+/// 2. checks whether `glue()` has exactly one unnamed argument with only `.open`
+///    and `.close` named arguments, and the string does not contain both specified
+///    delimiters (in which case using `.open` and `.close` is not needed).
```
etiennebacher:
```suggestion
/// 1. checks whether `glue()` evaluates some R code between delimiters;
/// 2. checks whether `glue()` would error when evaluated because of incomplete
///     delimiters.
```

Point 1 covers both `glue("x")` and `glue('{x}', .open = '<', .close = '>')`

crates/jarl-core/src/lints/base/glue/glue.rs
```diff
@@ -0,0 +1,207 @@
...
+///
+/// ## Why is this bad?
+///
+/// For 1, using `glue()` with only a constant string, e.g. `glue("abc")`, is
+/// useless and less readable. You can just use the string directly.
+///
+/// For 2, specifying `.open` and `.close` delimiters when the string does not
+/// contain those delimiters means `glue()` will not perform any interpolation,
+/// making the function call unnecessary.
```
etiennebacher: Point 2 should be that finding incomplete delimiters would error at runtime, cf this comment: https://github.com/etiennebacher/jarl/issues/331#issuecomment-4416077854

crates/jarl-core/src/lints/base/glue/glue.rs
```diff
@@ -0,0 +1,207 @@
...
+///
+/// Both cases do not have an automatic fix.
+///
+/// ## Example
+///
+/// ```r
+/// glue("abc")
+/// glue('{a}', .open = '<', .close = '>')
```
etiennebacher, line 33: This should have an example of e.g. `glue("{abc")`

crates/jarl-core/src/lints/base/glue/glue.rs
```diff
@@ -0,0 +1,207 @@
...
+    if fn_name != "glue" {
+        return Ok(None);
+    }
+    if let Some(ref ns) = fn_ns
+        && ns != "glue::"
+    {
+        return Ok(None);
+    }
+
+    // Don't know how to handle pipes for now.
```
etiennebacher: Can you add a TODO?

crates/jarl-core/src/lints/base/glue/glue.rs
```diff
@@ -0,0 +1,207 @@
...
+    let args = ast.arguments()?.items();
+    let named = get_named_args(&args);
+    let dots = get_unnamed_args(&args);
+
+    // If there is not exactly one unnamed argument, then we can't determine whether
+    // `glue()` is being used just to wrap a constant string or to concatenate
+    // multiple values, so skip linting.
+    if dots.len() != 1 {
+        return Ok(None);
+    }
```
etiennebacher: I don't understand this, why can't we determine that `glue("a", "b")` should be reported?

I suppose there are two approaches:

- concatenate unnamed arguments by hand and then analyze the concatenated content
-  loop over the unnamed arguments and report only if none evaluates code

I don't know which one is best.

crates/jarl-core/src/lints/base/glue/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+          | -------------------------------------- Using glue() with .open and .close when the string does not contain the specified delimiters performs no interpolation.
+          |
+        Found 1 error.
+        "
+        );
+    }
+
+    #[test]
+    fn test_no_lint_glue() {
+        expect_no_lint("glue('<a}', .open = '<')", "glue", None);
```
etiennebacher, line 67: This should report because it errors at runtime

etiennebacher (changes requested):
Thanks, I think this is on the right way but I have several comments to address below.

Also, please check the contributing guide for steps regarding the documentation (at the time of writing Netlify is down so the website isn't available but I guess this should be fixed soonish).

etiennebacher:
>  honestly have no idea what the codecov report is trying to say :)

Codecov fails when the PR decreases the overall coverage of the project

etiennebacher:
I think you also need to merge the main branch so that ecosystem results don't show that we lost violations (I shouldn't have rebased earlier, I think it only generated confusion, sorry).

crates/jarl-core/src/lints/base/glue/mod.rs
novica: I think it works, no?

```
R 4.5.1> a=1
R 4.5.1> glue::glue('<a}', .open = '<')
1
```

etiennebacher: Ah yes my bad

crates/jarl-core/src/lints/base/glue/glue.rs
novica: What do you mean by incomplete? This does not error:

`a=1`
`glue::glue('b <a}', .open="<")` 

Counting will not make much sense I think.

crates/jarl-core/src/lints/base/glue/glue.rs
novica: but this errors:
```
glue::glue("{abc")
Error in glue_data(.x = NULL, ..., .sep = .sep, .envir = .envir, .open = .open,  : 
  Expecting '}'
```

crates/jarl-core/src/lints/base/glue/glue.rs
novica: Just commenting out this check and running jarl will result in:

```
  |
2 | glue('a', 'b', 'c')
  | ------------------- glue() with a constant string performs no interpolation.
  |
 ```
 
 However,  I am not sure if `x <- "abc"` and ` x <- glue('a', 'b', 'c')` are the same thing? Maybe a user can have reasons to glue the strings?

**Comment by novica** [comment]
I am curious why do you prefer to manually link the PR to the issue instead of having the resolves-tag that does that automatically?

File: crates/jarl-core/src/lints/base/glue/glue.rs (hunk 2 above)
Comment (etiennebacher) [reply crates/jarl-core/src/lints/base/glue/glue.rs]: I mean `glue("{a")` for example, which would error at runtime but we can catch this early (before it actually runs). In the case of `glue('b <a}', .open="<")`, there is one left delimiter and one right delimiter (implicitly `}`) so we don't want to report this case.

File: crates/jarl-core/src/lints/base/glue/glue.rs (hunk 3 above)
Comment (etiennebacher) [reply crates/jarl-core/src/lints/base/glue/glue.rs:33]: Yes, we want to catch this early on.

File: crates/jarl-core/src/lints/base/glue/glue.rs (hunk 5 above)
Comment (etiennebacher) [reply crates/jarl-core/src/lints/base/glue/glue.rs]: > Maybe a user can have reasons to glue the strings?

Hmm true, maybe it's easier to read to break a long string in pieces and put those pieces in `glue()`. Ok, let's leave this unchanged, but can you update the comment to explain that this might be a conscious decision by the user?

**Comment by etiennebacher** [comment]
> I am curious why do you prefer to manually link the PR to the issue instead of having the resolves-tag that does that automatically?

I just prefer having the "Resolves #issue" or "Fix #issue" in the PR body rather than in its title because it's not particularly useful when I read the commit messages on the main branch on Github. I don't understand what you mean by "manually link", didn't you specify the issue number in the original title?

**Comment by novica** [comment]
Ok I think this turned out ok. current local test file is (not in the jarl tests):

```
  1 glue::glue("{aaa}") # ok
  2 glue('a', 'b', 'c') # ok.
  3 dlue('<a>', .open = '<', .close = '>') # ok
  4 glue::glue('{a>', 'b',  .close = ">") # ok
  5 glue('a') # report
  6 glue::glue("{abc") # report
  7 glue::glue('a', .sep = ' ') # report
  8 glue('{a}', .open = '<', .close = '>') # report
 ```

crates/jarl-core/src/lints/base/glue/glue.rs
```diff
@@ -0,0 +1,258 @@
...
+///    delimiters.
+///
+/// ## Why is this bad?
+///
+/// For 1, using `glue()` with only a constant string, e.g. `glue("abc")`, is
+/// useless and less readable. You can just use the string directly.
+///
+/// For 2, specifying `.open` and `.close` delimiters when the string does not
+/// contain those delimiters means `glue()` will not perform any interpolation,
+/// making the function call unnecessary.
```
etiennebacher:
```suggestion
/// For 2, having incomplete delimiters would error when evaluated,
/// so this indicates a bug.
```

crates/jarl-core/src/lints/base/glue/glue.rs
```diff
@@ -0,0 +1,258 @@
...
+/// glue("abc")
+/// glue('{a}', .open = '<', .close = '>')
+/// glue("{abc")
+/// ```
+///
+/// Use instead:
+/// ```r
+/// "abc"
+/// # For the second case, either use default delimiters {}, or ensure the string contains the specified delimiters
+/// # For the third case, fix the string to have complete delimiters, e.g. glue("{abc}")
```
etiennebacher:
```suggestion
/// # For the second case, either use default delimiters {},
/// # or ensure the string contains the specified delimiters
/// # For the third case, fix the string to have complete delimiters, 
/// # e.g. glue("{abc}")
```

crates/jarl-core/src/lints/base/glue/glue.rs
```diff
@@ -0,0 +1,258 @@
...
+pub fn glue(ast: &RCall) -> anyhow::Result<Option<Diagnostic>> {
+    let fn_name = get_function_name(ast.function()?);
+    let fn_ns = get_function_namespace_prefix(ast.function()?);
+
+    // Only trigger on `glue()` or `glue::glue()`
+    if fn_name != "glue" {
+        return Ok(None);
+    }
+    if let Some(ref ns) = fn_ns
+        && ns != "glue::"
```
etiennebacher, line 58: Can you try to increase the coverage? This line is apparently not covered, it would be with something like `foo::glue("abc")`

crates/jarl-core/src/lints/base/glue/glue.rs
```diff
@@ -0,0 +1,258 @@
...
+    let open_arg = get_arg_by_name(&args, ".open");
+    let close_arg = get_arg_by_name(&args, ".close");
+    let sep_arg = get_arg_by_name(&args, ".sep");
+    let open_text = get_named_string_arg_text(&args, ".open")?;
+    let close_text = get_named_string_arg_text(&args, ".close")?;
+    let sep_only = sep_arg.is_some();
+
+    if (open_arg.is_some() && open_text.is_none()) || (close_arg.is_some() && close_text.is_none())
+    {
+        return Ok(None);
```
etiennebacher, line 94: I think this could be tested with `glue("x", .close = )` (this and the previous comment are just examples to increase coverage, please refer to Codecov report if you have access to it from the Codecov comment above).

crates/jarl-core/src/lints/base/glue/glue.rs
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    let open = open_text.as_deref().unwrap_or("{");
+    let close = close_text.as_deref().unwrap_or("}");
+
+    let diagnostic = if has_incomplete_delimiters(&dot_text, open, close) {
+        Some(Diagnostic::new(
+            ViolationData::new(
+                "glue".to_string(),
+                "glue() contains incomplete delimiters and would error when evaluated.".to_string(),
```
etiennebacher:
```suggestion
                "`glue()` contains incomplete delimiters and would error when evaluated.".to_string(),
```

Same thing for all function calls and all arguments names in diagnostic messages

crates/jarl-core/src/lints/base/glue/glue.rs
```diff
@@ -0,0 +1,258 @@
...
+                None,
+            ),
+            ast.syntax().text_trimmed_range(),
+            Fix::empty(),
+        ))
+    } else if !dot_text.contains("{") && !dot_text.contains("}") && (named.is_empty()) {
```
etiennebacher:
```suggestion
    } else if !dot_text.contains("{") && !dot_text.contains("}") && open_arg.is_none() && close_arg.is_none() {
```

This looks more robust to me because it also captures cases where there are named args but not `.open` and `.close`

crates/jarl-core/src/lints/base/glue/glue.rs
```diff
@@ -0,0 +1,258 @@
...
+    } else if sep_only {
+        Some(Diagnostic::new(
+            ViolationData::new(
+                "glue".to_string(),
+                "glue() with a one constant string and .sep argument does not perform interpolation.".to_string(),
+                None,
+            ),
+            ast.syntax().text_trimmed_range(),
+            Fix::empty(),
+        ))
```
etiennebacher: I'm not convinced this branch is useful, it could be the same for all individual arguments apart from the delimiters.

crates/jarl-core/src/lints/base/glue/glue.rs
```diff
@@ -0,0 +1,258 @@
...
+    } else if let (Some(open), Some(close)) = (open_text, close_text) {
+        if !open.is_empty()
+            && !close.is_empty()
+            && (!dot_text.contains(&open) || !dot_text.contains(&close))
+        {
+            Some(Diagnostic::new(
+                ViolationData::new(
+                    "glue".to_string(),
+                    "Using glue() with .open and .close when the string does not contain the specified delimiters performs no interpolation.".to_string(),
```
etiennebacher:
```suggestion
                    "This `glue()` call isn't necessary because it performs no interpolation.".to_string(),
```
This feels clearer to me.

docs/rules.qmd
```diff
@@ -49,94 +49,105 @@ The column "Has fix" can take the following values:
...
+        `Rule name` = colDef(cell = function(value) {
+            htmltools::tags$a(
+                href = paste0("rules/", value, ".html"),
+                target = "_blank",
+                value
+            )
+        })
+    ),
+    defaultPageSize = 1000,
+    filterable = TRUE
```
etiennebacher: Can you remove formatting changes? (I should probably have an `air.toml` for that)

etiennebacher (changes requested):
Thanks, this looks better. I have some comments below. In addition to them, it seems that the violation reported in `arrow` is a false positive because a doubled delimiter is ignored by `glue()`, e.g. this is valid code:

```r
glue("<<x", .open = "<")
```
but this isn't:

```r
glue("<x", .open = "<")
```

crates/jarl-core/src/lints/base/glue/glue.rs
novica: I don't think it is the same for all. For example:
```
R 4.6.0> x = 1
R 4.6.0> glue::glue("
+    {
+      x + 1  # this is a comment
+    }
+  ", .comment = "#") 
2
```

crates/jarl-core/src/lints/base/glue/glue.rs
novica: OK. I think I solved this.

novica:
I think I messed up the clippy stuff by trying to have similar messages in the diagnostic.

crates/jarl-core/src/lints/base/glue/glue.rs
novica: I think this is the last thing we have open in this PR discussion.

etiennebacher:
@novica I've applied some fixes and simplifications myself because I had a hard time expressing myself properly in review comments :smile: let me know if you're fine with those changes before I merge

novica:
The main thing is you removed the `.sep` part. I have no objections. I guess it can be added again in the future if it turns out it is needed.

PR #487, opened 2026-05-19
fix: Skip chained assignments in `implicit_assignment`

### LINKED ISSUE #480: `implicit_assignment` does not flag chained assignments inside braced if blocks
I am not sure whether that's a bug or a feature, but I was surprised by the `jarl` behavior. The `implicit_assignment` rule flags chained assignments in unbraced `if` bodies, but silently ignores the exact same pattern once braces are added. 

For example:

```r
# 1. Unbraced if body — flagged
if (TRUE) a <- b <- 1
# 2. Braced if body — NOT flagged
if (TRUE) {
  a <- b <- 1
}
# 3. Nested braced blocks — NOT flagged
if (TRUE) {
  if (TRUE) {
    obj$x <- y <- foo()
  }
}
```

From my POV that's a bit inconsistent as braces don't make the assignment any less implicit.

atsyplenkov:
fix #480

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 ### Bug fixes
 
 * `implicit_assignment` no longer flags chained assignments like
-  `if (TRUE) a <- b <- 1`, aligning with `lintr` behavior (#480).
+  `if (TRUE) a <- b <- 1`, aligning with `lintr` behavior (#480, @atsyplenkov).
 
 * Automatic fixes now skip rewrites when comments are inside the rewritten
   expression, while still allowing leading and trailing comments (#461, @Yousa-Mirage).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@
 
 ### Bug fixes
 
+* `implicit_assignment` no longer flags chained assignments like
+  `if (TRUE) a <- b <- 1`, aligning with `lintr` behavior (#480).
+
 * Automatic fixes now skip rewrites when comments are inside the rewritten
   expression, while still allowing leading and trailing comments (#461, @Yousa-Mirage).
 
--- a/crates/jarl-core/src/lints/base/implicit_assignment/implicit_assignment.rs
+++ b/crates/jarl-core/src/lints/base/implicit_assignment/implicit_assignment.rs
@@ -57,6 +57,44 @@ pub fn implicit_assignment(
         return Ok(None);
     };
 
+    // Skip chained assignments like `a <- b <- 1` or `x <- y <<- z`
+
+    // Helper to check if an operator is an assignment operator
+    let is_assignment_op = |op: &RSyntaxToken| {
+        op.kind() == RSyntaxKind::ASSIGN
+            || op.kind() == RSyntaxKind::SUPER_ASSIGN
+            || op.kind() == RSyntaxKind::ASSIGN_RIGHT
+            || op.kind() == RSyntaxKind::SUPER_ASSIGN_RIGHT
+    };
+
+    // Early return: if this assignment's RHS is also an assignment
+    // (i.e., we're the left part of a chain like `a <- b <- 1`)
+    if let Ok(right) = ast.right() {
+        if let Some(right_binary) = right.as_r_binary_expression() {
+            if let Ok(right_op) = right_binary.operator() {
+                if is_assignment_op(&right_op) {
+                    return Ok(None);
+                }
+            }
+        }
+    }
+
+    // Early return: if this assignment is the RHS of a parent assignment
+    // (i.e., we're the right part of a chain like `a <- b <- 1`)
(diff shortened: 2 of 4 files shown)


crates/jarl-core/src/lints/base/implicit_assignment/implicit_assignment.rs
```diff
@@ -57,6 +57,38 @@ pub fn implicit_assignment(
...
+
+    // Early return: if this assignment's RHS is also an assignment
+    // (i.e., we're the left part of a chain like `a <- b <- 1`)
+    if let Ok(right) = ast.right()
+        && let Some(right_binary) = right.as_r_binary_expression()
+        && let Ok(right_op) = right_binary.operator()
+        && is_assignment_op(&right_op)
+    {
+        return Ok(None);
+    }
```
etiennebacher: Is this actually needed? In `if (foo) a <- b <- 1`, only `b <- 1` is highlighted by `implicit_assignment`, so do we need to handle the left part of the chain?

atsyplenkov:
Sorry, I initially missed the https://jarl.etiennebacher.com/contributing web page and my contribution wasnt really well formatted. It has been fixed in later commits.

In a nutshell, this small PR just adds two early returns when chains like this `a <- b <- 1` are present. And some tests to cover this (including triple chains like `a <- b <- c <- 1`)

docs/changelog.md
```diff
@@ -17,6 +17,9 @@
 
 ### Bug fixes
 
+* `implicit_assignment` no longer flags chained assignments like
+  `if (TRUE) a <- b <- 1`, aligning with `lintr` behavior (#480).
+
```
etiennebacher, line 22: You also need to add this in `CHANGELOG.md` (actually you could even modify `CHANGELOG.md` only and then `just document` will copy it to `docs/changelog.md`)

crates/jarl-core/src/lints/base/implicit_assignment/implicit_assignment.rs
```diff
@@ -57,6 +57,38 @@ pub fn implicit_assignment(
...
 
+    // Skip chained assignments like `a <- b <- 1` or `x <- y <<- z`
+
+    // Helper to check if an operator is an assignment operator
+    let is_assignment_op = |op: &RSyntaxToken| {
+        op.kind() == RSyntaxKind::ASSIGN
+            || op.kind() == RSyntaxKind::SUPER_ASSIGN
+            || op.kind() == RSyntaxKind::ASSIGN_RIGHT
+            || op.kind() == RSyntaxKind::SUPER_ASSIGN_RIGHT
+    };
```
etiennebacher: Can this be put higher so it can be used in the first if condition of the function?

etiennebacher (commented):
Thanks, looks good!

This doesn't fix #480 though, it will be closed when `chained_assignment` is implemented.

docs/changelog.md
atsyplenkov: Oh, nice! I have never used `just` before. Exciting

crates/jarl-core/src/lints/base/implicit_assignment/implicit_assignment.rs
atsyplenkov: Done

crates/jarl-core/src/lints/base/implicit_assignment/implicit_assignment.rs
atsyplenkov: Fixed

atsyplenkov:
Sorry for the delay with such small fix, I was distracted by my day job. I have addressed all your comments

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@
 
 ### Bug fixes
 
+* `implicit_assignment` no longer flags chained assignments like
+  `if (TRUE) a <- b <- 1`, aligning with `lintr` behavior (#480).
```
etiennebacher:
```suggestion
* `implicit_assignment` no longer flags chained assignments like
  `if (TRUE) a <- b <- 1`, aligning with `lintr` behavior (#480, @atsyplenkov).
```

docs/changelog.md
```diff
@@ -17,6 +17,9 @@
 
 ### Bug fixes
 
+* `implicit_assignment` no longer flags chained assignments like
+  `if (TRUE) a <- b <- 1`, aligning with `lintr` behavior (#480).
```
etiennebacher:
```suggestion
* `implicit_assignment` no longer flags chained assignments like
  `if (TRUE) a <- b <- 1`, aligning with `lintr` behavior (#480, @atsyplenkov).
```

etiennebacher (approved):
No worries at all, I certainly don't expect anyone to address comments immediately, I'm also not always available :)

Thanks again! I'm just adding your github username in the changelog, similar to other  contributors

PR #510, opened 2026-06-09
feat: Add `unnecessary_parentheses` rule

### LINKED ISSUE #463: Find unnecessary parenthesis
> There could be a new rule to find unnecessary parenthesis (for reference, this is reported by Ruff with code UP034), so that's not related to the `%notin%` rule. I think it's just a matter of checking if the first and only child of a `R_PARENTHESIZED_EXPRESSION` is also a `R_PARENTHESIZED_EXPRESSION`. 

 _Originally posted by @etiennebacher in [#239](https://github.com/etiennebacher/jarl/issues/239#issuecomment-4321957285)_

JosephBARBIERDARNAL:
Fixes #463


JosephBARBIERDARNAL:
Interesting, there are few valid unnecessary parenthesis in the ecosystem checks

JosephBARBIERDARNAL:
At the moment it raises one warning per unnecessary parenthesis per expression. For example, the following code:

```R
((x)) # LINT: 1
(((x))) # LINT: 2
((((x)))) # LINT: 3
```

will raise the following:

```R
warning: [unnecessary_parenthesis](https://jarl.etiennebacher.com/rules/unnecessary_parenthesis)
  --> temp.R:15:1
   |
15 | ((x))                         # LINT: 1
   | ----- This expression contains unnecessary parentheses.
   |
   = help: Remove one pair of parentheses.

warning: [unnecessary_parenthesis](https://jarl.etiennebacher.com/rules/unnecessary_parenthesis)
  --> temp.R:16:1
   |
16 | (((x)))                       # LINT: 2
   | ------- This expression contains unnecessary parentheses.
   |
   = help: Remove one pair of parentheses.

warning: [unnecessary_parenthesis](https://jarl.etiennebacher.com/rules/unnecessary_parenthesis)
  --> temp.R:16:2
   |
16 | (((x)))                       # LINT: 2
   |  ----- This expression contains unnecessary parentheses.
   |
   = help: Remove one pair of parentheses.

warning: [unnecessary_parenthesis](https://jarl.etiennebacher.com/rules/unnecessary_parenthesis)
  --> temp.R:17:1
   |
17 | ((((x))))                     # LINT: 3
   | --------- This expression contains unnecessary parentheses.
   |
   = help: Remove one pair of parentheses.

warning: [unnecessary_parenthesis](https://jarl.etiennebacher.com/rules/unnecessary_parenthesis)
  --> temp.R:17:2
   |
17 | ((((x))))                     # LINT: 3
   |  ------- This expression contains unnecessary parentheses.
   |
   = help: Remove one pair of parentheses.

warning: [unnecessary_parenthesis](https://jarl.etiennebacher.com/rules/unnecessary_parenthesis)
  --> temp.R:17:3
   |
17 | ((((x))))                     # LINT: 3
   |   ----- This expression contains unnecessary parentheses.
   |
   = help: Remove one pair of parentheses.
```

Maybe a better behavior would be to raise once and explicit the number of unnecesary parenthesis instead. Something like `This expression contains XX unnecessary parentheses`. But in practice I don't expect people to have multiple unnecesary parenthesis very often, so it might not matter.

JosephBARBIERDARNAL:
@etiennebacher This is a first draft for #463. Let me know what you think! I quickly checked for code on Github with the search you suggested and couldn't find cases that aren't properly handled, but I guess it's easy to miss them.

crates/jarl-core/src/lints/base/unnecessary_parenthesis/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+            "((x)) + y",
+            "if (((x))) y",
+            "(\n  (x)\n)",
+            "(\n  # explain x\n  (x)\n)",
+        ] {
+            assert!(
+                !check_code(code, "unnecessary_parenthesis", None).is_empty(),
+                "Expected a lint for: {code}",
+            );
+        }
```
etiennebacher: Please add a proper snapshot like  for each of these cases, like in the test above, and replace `\n` by actual line breaks.

crates/jarl-core/src/lints/base/unnecessary_parenthesis/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+    #[test]
+    fn test_reports_each_redundant_level() {
+        let diagnostics = check_code("(((x)))", "unnecessary_parenthesis", None);
+
+        assert_eq!(diagnostics.len(), 2);
+        assert_eq!(usize::from(diagnostics[0].range.start()), 0);
+        assert_eq!(usize::from(diagnostics[0].range.end()), 7);
+        assert_eq!(usize::from(diagnostics[1].range.start()), 1);
+        assert_eq!(usize::from(diagnostics[1].range.end()), 6);
+    }
```
etiennebacher: This type of test isn't necessary, snapshot tests also allow us to check visually that the reported range is correct.

crates/jarl-core/src/lints/base/unnecessary_parenthesis/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+            "(x)",
+            "(x + y) * z",
+            "foo(x)",
+            "if (x) y",
+            "while (x) y",
+            "function(x) x",
+        ] {
+            expect_no_lint(code, "unnecessary_parenthesis", None);
+        }
+    }
```
etiennebacher: Please make one `expect_no_lint()` per case

crates/jarl-core/src/lints/base/unnecessary_parenthesis/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    #[test]
+    fn test_unnecessary_parenthesis_metadata() {
+        assert!(Rule::UnnecessaryParenthesis.is_enabled_by_default());
+        assert!(Rule::UnnecessaryParenthesis.has_no_fix());
+    }
```
etiennebacher: No need for this

crates/jarl-core/src/lints/base/unnecessary_parenthesis/unnecessary_parenthesis.rs
```diff
@@ -0,0 +1,55 @@
...
+/// Version added: 0.6.0
+///
+/// ## What it does
+///
+/// Checks for expressions wrapped in multiple pairs of parentheses.
+///
+/// ## Why is this bad?
+///
+/// Repeated parentheses do not change the meaning of the expression and make
+/// the code harder to read.
```
etiennebacher:
```suggestion
/// Repeated parentheses do not change the meaning of the expression and can make
/// the code harder to read.
```

I try to be less definitive when talking about rules for "readability" only (but maybe I'm inconsistent about this in the documentation).

crates/jarl-core/src/lints/base/unnecessary_parenthesis/unnecessary_parenthesis.rs
```diff
@@ -0,0 +1,55 @@
...
+/// ```r
+/// (x + 1)
+/// ```
+impl Violation for UnnecessaryParenthesis {
+    fn name(&self) -> String {
+        "unnecessary_parenthesis".to_string()
+    }
+
+    fn body(&self) -> String {
+        "This expression contains unnecessary parentheses.".to_string()
```
etiennebacher: I never remember which is the preferred one between "parenthesis" and "parentheses" but it should be harmonized.

crates/jarl-core/src/lints/base/unnecessary_parenthesis/unnecessary_parenthesis.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    fn suggestion(&self) -> Option<String> {
+        Some("Remove one pair of parentheses.".to_string())
```
etiennebacher: Same

crates/jarl-core/src/rule_set.rs
```diff
@@ -615,6 +615,13 @@ declare_rules! {
         fix: Safe,
         min_r_version: None,
     },
+    UnnecessaryParenthesis => {
+        name: "unnecessary_parenthesis",
+        categories: [Read],
+        default: Enabled,
+        fix: None,
```
etiennebacher: We could have a safe fix to remove unnecessary parenthesis.

docs/_quarto.yml
```diff
@@ -152,7 +152,8 @@ website:
       - rules/unexplained_suppression.md
       - rules/unmatched_range_suppression.md
       - rules/unnecessary_nesting.md
+      - rules/unnecessary_parenthesis.md
       - rules/unreachable_code.md
       - rules/unused_function.md
       - rules/vector_logic.md
-      - rules/which_grepl.md
+      - rules/which_grepl.md
```
etiennebacher: Can you add a new line at the end of this file?

docs/changelog.md
```diff
@@ -16,6 +16,7 @@
   * `notin` (#459, @Yousa-Mirage)
   * `pipe_consistency` (#482)
   * `pipe_return` (#502)
+  * `unnecessary_parenthesis`
```
etiennebacher: Please add the PR number and your github username

docs/rules.qmd
```diff
@@ -131,6 +131,7 @@ dat <- as.data.frame(
     c("unexplained_suppression", "comments", "❌", ""),
     c("unmatched_range_suppression", "comments", "❌", ""),
     c("unnecessary_nesting", "readability", "✅", "Disabled by default"),
+    c("unnecessary_parenthesis", "readability", "❌", ""),
```
etiennebacher: Should be updated to show that a fix is available

etiennebacher (changes requested):
Thanks @JosephBARBIERDARNAL, this looks great!

> Maybe a better behavior would be to raise once and explicit the number of unnecesary parenthesis instead. Something like "This expression contains XX unnecessary parentheses."

I agree, I think the rule could have a `while` loop that counts the number of unnecessary parenthesis and reports a single violation. It's true it's unlikely to happen but let's make this rule as clean as possible from the start so that we don't have to come back to it later.

etiennebacher:
Small nitpick: please mark PRs as "ready to review" once you're happy with their state, otherwise I might just wait to review because it's still in draft mode :)

crates/jarl-core/src/lints/base/unnecessary_parenthesis/unnecessary_parenthesis.rs
JosephBARBIERDARNAL: Fair enough

crates/jarl-core/src/lints/base/unnecessary_parenthesis/unnecessary_parenthesis.rs
JosephBARBIERDARNAL: Same. Turns out it's just singular/plural. Still need to do some cleanup

crates/jarl-core/src/lints/base/unnecessary_parentheses/unnecessary_parentheses.rs
```diff
@@ -0,0 +1,82 @@
...
+impl Violation for UnnecessaryParentheses {
+    fn name(&self) -> String {
+        "unnecessary_parentheses".to_string()
+    }
+
+    fn body(&self) -> String {
+        format!(
+            "This expression contains {} unnecessary {} of parentheses.",
+            self.count,
+            if self.count == 1 { "pair" } else { "pairs" },
```
JosephBARBIERDARNAL: Would be fair to me to **not** explicit the count when it's 1 since this is what will happen most times (?) and isn't really relevant in those cases.

crates/jarl-core/src/lints/base/unnecessary_parentheses/unnecessary_parentheses.rs
```diff
@@ -0,0 +1,82 @@
...
+        )
+    }
+
+    fn suggestion(&self) -> Option<String> {
+        Some(format!(
+            "Remove {} {} of parentheses.",
+            self.count,
+            if self.count == 1 { "pair" } else { "pairs" },
```
JosephBARBIERDARNAL: same here

crates/jarl-core/src/rule_set.rs
JosephBARBIERDARNAL: I'll need to dig a bit more in the codebase to see how that works in practice.

docs/contributing-rule-example.md
```diff
@@ -16,7 +16,7 @@ So far, most (if not all) rules in Jarl come from [the list of rules available i
 If you want to add a rule that is not in `lintr`, please [open an issue](https://github.com/etiennebacher/jarl/issues/new/choose) first.
 
 Note that not all `lintr` rules are suitable for Jarl.
-In particular, rules that are only about formatting (spaces before parenthesis, newlines between arguments, etc.) are **out of scope** for Jarl.
+In particular, rules that are only about formatting (spaces before parentheses, newlines between arguments, etc.) are **out of scope** for Jarl.
```
JosephBARBIERDARNAL, line 19: While making changes related to parenthesis/parentheses, I found this one that should be in plural

crates/jarl-core/src/lints/base/unnecessary_parentheses/unnecessary_parentheses.rs
etiennebacher: Agreed

crates/jarl-core/src/lints/base/unnecessary_parentheses/unnecessary_parentheses.rs
```diff
@@ -0,0 +1,82 @@
...
+        ))
+    }
+}
```
etiennebacher: I see the appeal of having a `count` field in the `UnnecessaryParentheses` struct and using it to customize the message (honestly I hadn't thought about this approach and it looks nice), but in the other rules I customize the message and the suggestion in the body of the rule (here `unnecessary_parentheses()`). For consistency I think the same approach should be used here (meaning that `impl Violation for UnnecessaryParentheses {}` should be removed).

JosephBARBIERDARNAL: makes sense, just moved it inside the body!

crates/jarl-core/src/lints/base/unnecessary_parentheses/mod.rs
```diff
@@ -0,0 +1,196 @@
...
+                ],
+                "unnecessary_parentheses",
+                None,
+            )
+        );
+    }
+
+    #[test]
+    fn test_unnecessary_parentheses_with_comments_no_fix() {
+        // Should detect lint but skip fix when comments are present to avoid destroying them
```
JosephBARBIERDARNAL: I added a comment to explicit it here because that wasn't obvious to me at first, but we can remove it if prefered.

JosephBARBIERDARNAL:
Finally took some time to implement the fix option.

Current fix code is a bit verbose because we need to check for precedence. The precedence table I added isn't specific to this rule but we must know whether the last pair is semantically required. This:

```R
((x + y)) * z
```

shouldn't become:

```R
x + y * z
```

Right now only this rule needs it (?) so maybe it's fair to keep it here?

Note that this would also be possible (**and maybe better while I think about it**) to have something that doesn't remove as much unnecessary parentheses but keep code "much" simpler (e.g., remove most of the helpers in this file).

The only downside of doing this is that this:

```R
((x * y)) + z
```

would become:

```R
(x * y) + z
```

instead of (current behavior):

```R
x * y + z
```

The upside would be to remove some code (~120 lines) which I think is valuable. Let me know what you think!

Ready for another pass of review @etiennebacher

crates/jarl-core/src/lints/base/unnecessary_parentheses/unnecessary_parentheses.rs
```diff
@@ -0,0 +1,227 @@
...
+
+    if parent.right()?.syntax().text_trimmed_range() == range {
+        return Ok(Some(ChildSide::Right));
+    }
+
+    Ok(None)
+}
+
+// Returns the precedence level of a binary operator (higher = tighter binding), or None for unknown operators.
+fn binary_precedence(operator: RSyntaxKind) -> Option<u8> {
```
JosephBARBIERDARNAL: TBH I'm unsure if this table is complete/exhaustive, but AI says so. See other comment in the main thread.

etiennebacher (changes requested):
Thank you, I think we should keep this rule simple and remove the precedence analysis. I'd rather have this rule recommend (and autofix) `((x * y)) + z` to become `(x * y) + z` rather than `x * y + z` because:

1. adding precedence analysis makes the rule more complex and I'm a bit worried we apply wrong fixes to cases where this would have big consequences
2. sometimes you want to use parenthesis to clearly distinguish terms of a formula, even they don't have any effect on the precedence, so I'd rather not report those cases.

JosephBARBIERDARNAL:
@etiennebacher that makes sense, I just removed the precedence analysis. This is ready for review again!

etiennebacher:
Looks great @JosephBARBIERDARNAL, thank you!

PR #530, opened 2026-06-28
chore: Upgrade project dependencies

Goldziher:
## Context

I am building a multi-language linting tool and want to depend on `jarl` (and `air`) directly. The blocker was the pinned `tree-sitter` version: the `air` dependency held it at 0.24, which conflicts with the rest of my toolchain that needs tree-sitter 0.26+. To unblock this I updated `air` to build on a newer tree-sitter, then updated `jarl` to consume it.

This PR brings `jarl`'s dependencies up to date and lifts tree-sitter to 0.26.9.

## What changed

- `cargo upgrade --incompatible` plus `cargo update` across the workspace.
- The `air` git dependencies now point at a fork (`Goldziher/r-air-fork`) that builds on tree-sitter 0.26.9. This is the only way to move tree-sitter off 0.24 today, since it is a transitive dependency of `air`.
- `annotate-snippets` 0.11 to 0.12: `render_diagnostic` was rewritten for the new Group/Element API. The rendered diagnostic output is preserved (Context annotation, trailing Padding, in-group help footer), so user-facing output is unchanged.
- `lsp-types` 0.95 to 0.97: `Url` was replaced by `Uri` (fluent-uri), which has no file-path conversions. A small `uri_ext` helper round-trips through the `url` crate for those conversions.
- A few snapshots were updated where the new `annotate-snippets` line-folding shows slightly more context, and one toml parse-error message changed with `toml` 1.1.

## Status

This depends on the corresponding `air` change landing upstream first. Until `air` merges and a stable rev is available, the `air` dependency here points at the fork. Once `air` is merged I will repoint these dependencies at the upstream rev. Opening as a regular (non-draft) PR so CI runs against the current state.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -29,7 +29,7 @@ dependencies = [
 [[package]]
 name = "air_r_factory"
 version = "0.0.0"
-source = "git+https://github.com/posit-dev/air?rev=0eadd6938d0316b1fc885c91e7af55d7c8de8530#0eadd6938d0316b1fc885c91e7af55d7c8de8530"
+source = "git+https://github.com/Goldziher/r-air-fork?rev=c916545f14f76e1d6bd6ff918870f86dfa704b63#c916545f14f76e1d6bd6ff918870f86dfa704b63"
 dependencies = [
  "air_r_syntax",
  "biome_rowan",
@@ -38,13 +38,14 @@ dependencies = [
 [[package]]
 name = "air_r_formatter"
 version = "0.0.0"
-source = "git+https://github.com/posit-dev/air?rev=0eadd6938d0316b1fc885c91e7af55d7c8de8530#0eadd6938d0316b1fc885c91e7af55d7c8de8530"
+source = "git+https://github.com/Goldziher/r-air-fork?rev=c916545f14f76e1d6bd6ff918870f86dfa704b63#c916545f14f76e1d6bd6ff918870f86dfa704b63"
 dependencies = [
  "air_r_syntax",
  "biome_formatter",
  "biome_rowan",
  "comments",
  "itertools",
+ "line_ending",
  "settings",
  "tracing",
 ]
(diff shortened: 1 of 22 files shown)


etiennebacher:
Hi, thanks but a few points to discuss first:

- Jarl is meant to be used as binary and not as a library (for now at least), and therefore I don't do changes while taking into account the stability of the internal API. How do you plan to use Jarl? 
- some of these dependencies were not bumped by choice:
   https://github.com/etiennebacher/jarl/blob/a02901538aefe3749ae5e18e054b35be49cdc8d4/.github/dependabot.yml#L36-L43
  See for instance the bad perf impact of `annotate_snippets` on some projects.

- since this depends on your fork of Air, I'm moving this to draft

Goldziher:
> Hi, thanks but a few points to discuss first:
> 
> * Jarl is meant to be used as binary and not as a library (for now at least), and therefore I don't do changes while taking into account the stability of the internal API. How do you plan to use Jarl?
> * some of these dependencies were not bumped by choice:
>   https://github.com/etiennebacher/jarl/blob/a02901538aefe3749ae5e18e054b35be49cdc8d4/.github/dependabot.yml#L36-L43
>   
>   See for instance the bad perf impact of `annotate_snippets` on some projects.
> * since this depends on your fork of Air, I'm moving this to draft

Thanks. 

I updated the title to conventional commits. 

1. I use it as a git rev dependency, same as i use other suchlike binary tools. Its preferable to actually forking. It would of course be great if you decide to publish as a crate i can install.

2. I will address the pins, and revert as needed. But are you certain? Some include API changes I already did. 

Best

etiennebacher:
> I use it as a git rev dependency, same as i use other suchlike binary tools. Its preferable to actually forking. It would of course be great if you decide to publish as a crate i can install.

Sure, maybe at some point but for now I don't force myself to maintain a stable API since the target is binary users. 

> I will address the pins, and revert as needed. But are you certain?

`data.table` and `stringmagic` have large performance regression in the benchmark results above so I wouldn't merge this PR as-is even if the equivalent PR in Air was merged.

Goldziher:
Closing this — the dependency upgrade here has gone stale and now conflicts. Agreed that a published crate is the right path rather than a git rev dependency; I'll revisit once that's in place.

PR #546, opened 2026-07-10
feat: Implement `strings_as_factors` rule

Yousa-Mirage:
See the [`lintr` rule](https://lintr.r-lib.org/reference/strings_as_factors_linter.html)

Related to #8

### DIFF
--- a/crates/jarl-core/src/analyze/call.rs
+++ b/crates/jarl-core/src/analyze/call.rs
@@ -118,7 +118,7 @@ pub fn call(r_expr: &RCall, checker: &mut Checker) -> anyhow::Result<()> {
         checker.report_diagnostic(sprintf(r_expr)?);
     }
     if checker.is_rule_enabled(Rule::StringsAsFactors) {
-        checker.report_diagnostic(strings_as_factors(r_expr, checker.minimum_r_version)?);
+        checker.report_diagnostic(strings_as_factors(r_expr, checker)?);
     }
     if checker.is_rule_enabled(Rule::SystemFile) {
         checker.report_diagnostic(system_file(r_expr)?);
--- a/crates/jarl-core/src/analyze/call.rs
+++ b/crates/jarl-core/src/analyze/call.rs
@@ -26,6 +26,7 @@ use crate::lints::base::redundant_ifelse::redundant_ifelse::redundant_ifelse;
 use crate::lints::base::sample_int::sample_int::sample_int;
 use crate::lints::base::seq2::seq2::seq2;
 use crate::lints::base::sprintf::sprintf::sprintf;
+use crate::lints::base::strings_as_factors::strings_as_factors::strings_as_factors;
 use crate::lints::base::system_file::system_file::system_file;
 use crate::lints::base::undesirable_function::undesirable_function::undesirable_function;
 use crate::lints::base::which_grepl::which_grepl::which_grepl;
@@ -115,6 +116,9 @@ pub fn call(r_expr: &RCall, checker: &mut Checker) -> anyhow::Result<()> {
     }
     if checker.is_rule_enabled(Rule::Sprintf) {
         checker.report_diagnostic(sprintf(r_expr)?);
+    }
+    if checker.is_rule_enabled(Rule::StringsAsFactors) {
+        checker.report_diagnostic(strings_as_factors(r_expr, checker.minimum_r_version)?);
     }
     if checker.is_rule_enabled(Rule::SystemFile) {
         checker.report_diagnostic(system_file(r_expr)?);
--- a/crates/jarl-core/src/lints/base/strings_as_factors/mod.rs
+++ b/crates/jarl-core/src/lints/base/strings_as_factors/mod.rs
@@ -31,10 +31,40 @@ mod tests {
             "strings_as_factors",
             Some("3.6"),
         );
+        expect_no_lint(
+            "data.frame(x = \"a\", \"stringsAsFactors\" = FALSE)",
+            "strings_as_factors",
+            Some("3.6"),
+        );
+        expect_no_lint(
+            "data.frame(x = \"a\", 'stringsAsFactors' = FALSE)",
+            "strings_as_factors",
+            Some("3.6"),
+        );
+        expect_no_lint(
+            "data.frame(x = \"a\", `stringsAsFactors` = FALSE)",
+            "strings_as_factors",
+            Some("3.6"),
+        );
 
         expect_no_lint("data.frame(x = 1.2)", "strings_as_factors", Some("3.6"));
         expect_no_lint(
             "data.frame(row.names = \"a\")",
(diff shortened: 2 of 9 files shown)


Yousa-Mirage:
OK, ready for review.

crates/jarl-core/src/lints/base/strings_as_factors/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+        expect_no_lint(
+            "data.frame(a = c('b c' = 1))",
+            "strings_as_factors",
+            Some("3.6"),
+        );
+        expect_no_lint(
+            "data.frame(x = c(\"a\", y))",
+            "strings_as_factors",
+            Some("3.6"),
+        );
```
etiennebacher, line 65: Just a comment for myself since I thought that `c()` would always coerce everything to character if there's at least one character value, from `?c`:

> The output type is determined from the highest type of the components in the hierarchy NULL < raw < logical < integer < double < complex < character < list < expression. 

``` r
c("a", list(1))
#> [[1]]
#> [1] "a"
#> 
#> [[2]]
#> [1] 1
```

crates/jarl-core/src/lints/base/strings_as_factors/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+        expect_no_lint(
+            "data.frame(x = rep(, 2))",
+            "strings_as_factors",
+            Some("3.6"),
+        );
+        expect_no_lint(
+            "data.frame(x = rep(times = 2, x = \"a\"))",
+            "strings_as_factors",
+            Some("3.6"),
+        );
```
etiennebacher: I don't think those add anything compared to the test above

crates/jarl-core/src/lints/base/strings_as_factors/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+          | -------------------------------------- `data.frame()` can create different column types before and after R 4.0 when `stringsAsFactors` is omitted.
+          |
+          = help: Specify `stringsAsFactors = TRUE` or `stringsAsFactors = FALSE` explicitly.
+        Found 1 error.
+        "#
+        );
+    }
+
+    #[test]
+    fn test_lint_static_character_expressions() {
```
etiennebacher: Please add the rule name in all test functions

crates/jarl-core/src/lints/base/strings_as_factors/strings_as_factors.rs
```diff
@@ -0,0 +1,196 @@
...
+///
+/// Use one of the following instead:
+/// ```r
+/// data.frame(x = "a", stringsAsFactors = TRUE)
+/// data.frame(x = "a", stringsAsFactors = FALSE)
+/// ```
+///
+/// ## References
+///
+/// See the [`lintr` rule](https://lintr.r-lib.org/reference/strings_as_factors_linter.html)
```
etiennebacher: You can replace the link to `lintr` docs by `?data.frame` (just for consistency since I haven't linked to `lintr` docs elsewhere)

crates/jarl-core/src/analyze/call.rs
```diff
@@ -116,6 +117,9 @@ pub fn call(r_expr: &RCall, checker: &mut Checker) -> anyhow::Result<()> {
     if checker.is_rule_enabled(Rule::Sprintf) {
         checker.report_diagnostic(sprintf(r_expr)?);
     }
+    if checker.is_rule_enabled(Rule::StringsAsFactors) {
+        checker.report_diagnostic(strings_as_factors(r_expr, checker.minimum_r_version)?);
+    }
```
etiennebacher, line 122: Just for consistency with the other calls, can you pass the entire `checker` and then use `checker.minimum_r_version` in the function body?

crates/jarl-core/src/lints/base/strings_as_factors/strings_as_factors.rs
```diff
@@ -0,0 +1,196 @@
...
+    let RCallFields { function, arguments } = ast.as_fields();
+    if get_function_name(function?) != "data.frame" {
+        return Ok(None);
+    }
+
+    // Check if `stringsAsFactors` is explicitly set.
+    let arguments = arguments?.items();
+    if get_arg_by_name(&arguments, "stringsAsFactors").is_some() {
+        return Ok(None);
+    }
```
etiennebacher: I think you can replace that by `unwrap_or_return_none!()`

crates/jarl-core/src/lints/base/strings_as_factors/strings_as_factors.rs
```diff
@@ -0,0 +1,196 @@
...
+            .is_ok_and(|function| get_function_name(function) == "c")
+            && is_literal_character_combine(call)
+    })
+}
+
+fn is_string_literal(expression: &AnyRExpression) -> bool {
+    expression
+        .as_any_r_value()
+        .is_some_and(|value| value.as_r_string_value().is_some())
+}
```
etiennebacher, line 216: Does this handle raw strings?

crates/jarl-core/src/lints/base/strings_as_factors/strings_as_factors.rs
```diff
@@ -0,0 +1,196 @@
...
+    if is_string_literal(&expression) {
+        return true;
+    }
+
+    let Some(call) = expression.as_r_call() else {
+        return false;
+    };
+    let function = call
+        .function()
+        .expect("a call in a valid expression must have a function");
```
etiennebacher: Here and in other places I would prefer the function to return a `Result` instead of having `expect()` that could panic.

etiennebacher (changes requested):
Thank you, looks good, some comments to address below.

etiennebacher:
Can you also update `docs/rules.qmd`?

crates/jarl-core/src/lints/base/strings_as_factors/strings_as_factors.rs
Yousa-Mirage: I don't think so. Here, `Ok(None)` is returned if the value exists, but `unwrap_or_return_none!()` returns `Ok(None)` if the value does not exist.

crates/jarl-core/src/lints/base/strings_as_factors/strings_as_factors.rs
Yousa-Mirage: Yes, `as_r_string_value()` can handle raw strings. And the raw string is added into R after v4.0.0 meanwhile `strings_as_factors` rule is only enabled in versions of R prior to 4.0.0. So they won't meet.

etiennebacher (approved):

etiennebacher:
Thanks!

PR #547, opened 2026-07-10
feat: Implement `stopifnot_all` rule

Yousa-Mirage:
See the [`lintr` rule](https://lintr.r-lib.org/reference/stopifnot_all_linter.html)

Related to #8

### DIFF
--- a/crates/jarl-core/src/lints/base/stopifnot_all/mod.rs
+++ b/crates/jarl-core/src/lints/base/stopifnot_all/mod.rs
@@ -30,29 +30,84 @@ mod tests {
          --> <test>:1:11
           |
         1 | stopifnot(all(x > 0))
-          |           ---------- `stopifnot(all(x))` produces a less informative error message.
+          |           ---------- `stopifnot(all(...))` contains an unnecessary call to `all()`.
           |
-          = help: Use `stopifnot(x)` instead.
+          = help: Use `stopifnot(...)` instead.
         Found 1 error.
         "
         );
 
-        for code in [
-            "stopifnot(check = all(x))",
-            "base::stopifnot(base::all(x))",
-            "stopifnot(all(x, na.rm = TRUE))",
-            "stopifnot(all(x, y))",
-        ] {
-            assert_eq!(
-                check_code(code, "stopifnot_all", None).len(),
-                1,
-                "expected a lint for `{code}`"
-            );
-        }
-
-        assert_eq!(
-            check_code("stopifnot(x, all(y), all(z))", "stopifnot_all", None).len(),
-            2
+        assert_snapshot!(
+            snapshot_lint("stopifnot(check = all(x))"),
+            @r"
+        warning: stopifnot_all
+         --> <test>:1:19
+          |
+        1 | stopifnot(check = all(x))
+          |                   ------ `stopifnot(all(...))` contains an unnecessary call to `all()`.
+          |
+          = help: Use `stopifnot(...)` instead.
+        Found 1 error.
+        "
+        );
+        assert_snapshot!(
+            snapshot_lint("base::stopifnot(base::all(x))"),
+            @r"
+        warning: stopifnot_all
+         --> <test>:1:17
+          |
+        1 | base::stopifnot(base::all(x))
+          |                 ------------ `stopifnot(all(...))` contains an unnecessary call to `all()`.
+          |
+          = help: Use `stopifnot(...)` instead.
+        Found 1 error.
+        "
+        );
+        assert_snapshot!(
+            snapshot_lint("stopifnot(all(x, na.rm = TRUE))"),
+            @r"
+        warning: stopifnot_all
+         --> <test>:1:11
+          |
+        1 | stopifnot(all(x, na.rm = TRUE))
+          |           -------------------- `stopifnot(all(...))` contains an unnecessary call to `all()`.
+          |
+          = help: Use `stopifnot(...)` instead.
+        Found 1 error.
+        "
+        );
+        assert_snapshot!(
+            snapshot_lint("stopifnot(all(x, y))"),
+            @r"
+        warning: stopifnot_all
+         --> <test>:1:11
+          |
+        1 | stopifnot(all(x, y))
+          |           --------- `stopifnot(all(...))` contains an unnecessary call to `all()`.
+          |
+          = help: Use `stopifnot(...)` instead.
+        Found 1 error.
+        "
+        );
+        assert_snapshot!(
+            snapshot_lint("stopifnot(x, all(y), all(z))"),
+            @r"
+        warning: stopifnot_all
+         --> <test>:1:14
+          |
+        1 | stopifnot(x, all(y), all(z))
+          |              ------ `stopifnot(all(...))` contains an unnecessary call to `all()`.
+          |
+          = help: Use `stopifnot(...)` instead.
+        warning: stopifnot_all
+         --> <test>:1:22
+          |
+        1 | stopifnot(x, all(y), all(z))
+          |                      ------ `stopifnot(all(...))` contains an unnecessary call to `all()`.
+          |
+          = help: Use `stopifnot(...)` instead.
+        Found 2 errors.
+        "
         );
     }
 }
--- a/crates/jarl-core/src/lints/base/stopifnot_all/mod.rs
+++ b/crates/jarl-core/src/lints/base/stopifnot_all/mod.rs
@@ -1,0 +1,58 @@
+pub(crate) mod stopifnot_all;
+
+#[cfg(test)]
+mod tests {
+    use crate::utils_test::*;
+    use insta::assert_snapshot;
+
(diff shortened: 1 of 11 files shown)


Yousa-Mirage:
OK, ready for review.

crates/jarl-core/src/lints/base/stopifnot_all/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+                check_code(code, "stopifnot_all", None).len(),
+                1,
+                "expected a lint for `{code}`"
+            );
+        }
+
+        assert_eq!(
+            check_code("stopifnot(x, all(y), all(z))", "stopifnot_all", None).len(),
+            2
+        );
```
etiennebacher: Can you expand these assertions to show the entire snapshots? There are not too many cases here

crates/jarl-core/src/lints/base/stopifnot_all/stopifnot_all.rs
```diff
@@ -0,0 +1,72 @@
...
+///
+/// ## What it does
+///
+/// Checks for direct calls to `all()` inside `stopifnot()`.
+///
+/// ## Why is this bad?
+///
+/// `stopifnot()` already checks `all()` of each argument internally. Passing
+/// `all(x)` hides the original expression from `stopifnot()`, which results in
+/// a less informative error message when the condition fails.
```
etiennebacher: I'm not convinced by the argument about the error message:

``` r
x <- -1:1
stopifnot(all(x > 0))
#> Error:
#> ! all(x > 0) is not TRUE
stopifnot(x > 0)
#> Error:
#> ! x > 0 are not all TRUE
```

I think we can just say it removes an unnecessary call.

crates/jarl-core/src/lints/base/stopifnot_all/stopifnot_all.rs
```diff
@@ -0,0 +1,72 @@
...
+///
+/// Use instead:
+/// ```r
+/// stopifnot(x > 0)
+/// ```
+///
+/// ## References
+///
+/// See the [`lintr` rule](https://lintr.r-lib.org/reference/stopifnot_all_linter.html)
+/// and `?stopifnot`.
```
etiennebacher:
```suggestion
/// See `?stopifnot`.
```

crates/jarl-core/src/lints/base/stopifnot_all/stopifnot_all.rs
```diff
@@ -0,0 +1,72 @@
...
+impl Violation for StopifnotAll {
+    fn name(&self) -> String {
+        "stopifnot_all".to_string()
+    }
+
+    fn body(&self) -> String {
+        "`stopifnot(all(x))` produces a less informative error message.".to_string()
```
etiennebacher:
```suggestion
        "`stopifnot(all(...))` contains an unnecessary call to `all()`.".to_string()
```

crates/jarl-core/src/lints/base/stopifnot_all/stopifnot_all.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    fn suggestion(&self) -> Option<String> {
+        Some("Use `stopifnot(x)` instead.".to_string())
```
etiennebacher:
```suggestion
        Some("Use `stopifnot(...)` instead.".to_string())
```

crates/jarl-core/src/lints/base/stopifnot_all/stopifnot_all.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+pub fn stopifnot_all(ast: &RCall) -> anyhow::Result<Option<Diagnostic>> {
+    if get_function_name(ast.function()?) != "all" {
+        return Ok(None);
+    }
+
+    let Some(argument) = ast.syntax().parent().and_then(RArgument::cast) else {
+        return Ok(None);
+    };
```
etiennebacher: You can use `unwrap_or_return_none!()`

crates/jarl-core/src/lints/base/stopifnot_all/stopifnot_all.rs
```diff
@@ -0,0 +1,72 @@
...
+    let outer_call = argument
+        .syntax()
+        .ancestors()
+        .find_map(RCall::cast)
+        .expect("an R argument must belong to a call");
```
etiennebacher: Can we have this return a `Result` instead of potentially panicking?

crates/jarl-core/src/lints/base/stopifnot_all/stopifnot_all.rs
```diff
@@ -0,0 +1,72 @@
...
+    fn body(&self) -> String {
+        "`stopifnot(all(x))` produces a less informative error message.".to_string()
+    }
+
+    fn suggestion(&self) -> Option<String> {
+        Some("Use `stopifnot(x)` instead.".to_string())
+    }
+}
+
+pub fn stopifnot_all(ast: &RCall) -> anyhow::Result<Option<Diagnostic>> {
```
etiennebacher, line 53: Worth adding a comment somewhere that this approach is a bit different from the other nested function calls because `all()` can be anywhere in `stopifnot()`, so we need to start from `all()` and check the parent.

crates/jarl-core/src/lints/base/stopifnot_all/stopifnot_all.rs
```diff
@@ -0,0 +1,72 @@
...
+        .expect("an R argument must belong to a call");
+
+    if get_function_name(outer_call.function()?) != "stopifnot" {
+        return Ok(None);
+    }
+
+    Ok(Some(Diagnostic::new(
+        StopifnotAll,
+        ast.syntax().text_trimmed_range(),
+        Fix::empty(),
```
etiennebacher: I think we can have an automatic fix that just removes the `all()`.

crates/jarl-core/src/rule_set.rs
```diff
@@ -594,6 +594,13 @@ declare_rules! {
         fix: Safe,
         min_r_version: None,
     },
+    StopifnotAll => {
+        name: "stopifnot_all",
+        categories: [Read],
+        default: Enabled,
```
etiennebacher: I think it's better to have this disabled by default, it's really not critical and doesn't harm readability much (also needs to be updated in the docs).

docs/_quarto.yml
```diff
@@ -147,6 +147,7 @@ website:
       - rules/seq2.md
       - rules/sort.md
       - rules/sprintf.md
+      - rules/stopifnot_all.md
```
etiennebacher, line 150: can you also update `docs/rules.qmd`?

etiennebacher (changes requested):
Thank you, see some comments below.

crates/jarl-core/src/lints/base/stopifnot_all/stopifnot_all.rs
Yousa-Mirage: Removing `all()` may cause some problems:

```r
stopifnot(all(c(TRUE, NA), na.rm = TRUE)) # pass
stopifnot(c(TRUE, NA), na.rm = TRUE)
#> Error:
#> ! c(TRUE, NA) are not all TRUE

stopifnot(all(1, 2, 3)) # pass but warning
#> Warning in all(1, 2, 3): coercing argument of type 'double' to logical
#> Warning in all(1, 2, 3): coercing argument of type 'double' to logical
#> Warning in all(1, 2, 3): coercing argument of type 'double' to logical
stopifnot(1, 2, 3)
#> Error:
#> ! 1 is not TRUE
```

Do you think we should provide an automatic unsafe fix?

etiennebacher: In this case we shouldn't report calls that contain `na.rm = TRUE` because the user wouldn't be able to do anything about that.

The example with coercion is a valid case where we would break code, so let's make the fix unsafe, yes.

etiennebacher: We can also mention in docs that this rule has an automatic unsafe fix.

docs/changelog.md
```diff
@@ -18,6 +18,7 @@
   * `notin` (#459, @Yousa-Mirage)
   * `pipe_consistency` (#482)
   * `pipe_return` (#502)
+  * `stopifnot_all` (@Yousa-Mirage)
```
etiennebacher: Please add the PR number (also for `strings_as_factors`)

Yousa-Mirage:
OK. I've added an unsafe fix and a test for `stopifnot(all(c(TRUE, NA), na.rm = TRUE))`. Please review it. Thank you for your detailed comments.

etiennebacher (approved):
Thanks!

PR #548, opened 2026-07-10
fix: Clippy warnings in Rust 1.97.0

Yousa-Mirage:
As the title, this PR fixed some new clippy warnings in Rust 1.97.0


etiennebacher:
Thanks!

Issue #232, opened 2025-12-08
`expect_identical(class(x), <any>)` should be fixable


DavisVaughan:
To `expect_s3_class()`

This is particularly helpful since it would make it easier for us as package maintainers to change our package internals if less people are relying on them due to jarl pushing them away from this

hadley:
In that case you can use `expect_s3_class(..., exact = TRUE)` 😄

etiennebacher:
Right, I'll mention this in the docs, thanks!

PR #553, opened 2026-07-17
feat: Implement `expect_s4_class_linter`

Yousa-Mirage:
Part of #8

### DIFF
--- a/crates/jarl-core/src/analyze/call.rs
+++ b/crates/jarl-core/src/analyze/call.rs
@@ -42,6 +42,7 @@ use crate::lints::testthat::expect_no_match::expect_no_match::expect_no_match;
 use crate::lints::testthat::expect_not::expect_not::expect_not;
 use crate::lints::testthat::expect_null::expect_null::expect_null;
 use crate::lints::testthat::expect_s3_class::expect_s3_class::expect_s3_class;
+use crate::lints::testthat::expect_s4_class::expect_s4_class::expect_s4_class;
 use crate::lints::testthat::expect_true_false::expect_true_false::expect_true_false;
 use crate::lints::testthat::expect_type::expect_type::expect_type;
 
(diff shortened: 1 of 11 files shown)


etiennebacher (approved):
Thanks!

Worth noting that some autofixes may fail but these would indicate that the original test is wrongly specified anyway so I don't think this warrants marking the fix as unsafe: https://github.com/r-lib/lintr/issues/2987

Yousa-Mirage:
Thank you for reminding me of this. I didn't notice this problem at all before. I think this is why `expect_s4_class_linter` has never been implemented. But I agree with you, in these cases `is()` should not be used in the first place.

I'd' like to discuss with you here the issue of `expect_s3_class()`, which was discussed in #232. When implementing S4's linter, I found three areas of the `expect_s3_class` rule that can be optimized:
1. When discovering the pattern `expect_identical(class(x), "class")`, should we proactively add `exact = TRUE`? Because `expect_identical()` may express an exact class matching intent.
2. [`lintr::expect_s3_class_linter()`](https://github.com/r-lib/lintr/blob/main/R/expect_s3_class_linter.R) will check some other modes not mentioned in the documentation:
	``` r
	library(lintr)
	
	lint(
	  text = 'expect_true(inherits(x, "foo"))',
	  linters = expect_s3_class_linter()
	)
	#> <text>:1:1: warning: [expect_s3_class_linter] expect_s3_class(x, k) is better than expect_true(is.<k>(x)) or expect_true(inherits(x, k)). Note also expect_s4_class() available for testing S4 objects.
	#> expect_true(inherits(x, "foo"))
	#> ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
	
	lint(
	  text = 'expect_true(is.data.frame(x))',
	  linters = expect_s3_class_linter()
	)
	#> <text>:1:1: warning: [expect_s3_class_linter] expect_s3_class(x, k) is better than expect_true(is.<k>(x)) or expect_true(inherits(x, k)). Note also expect_s4_class() available for testing S4 objects.
	#> expect_true(is.data.frame(x))
	#> ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
	
	lint(
	  text = 'expect_equal(class(x), classes)',
	  linters = expect_s3_class_linter()
	)
	#> <text>:1:1: warning: [expect_s3_class_linter] expect_s3_class(x, k) is better than expect_equal(class(x), k). Note also expect_s4_class() available for testing S4 objects.
	#> expect_equal(class(x), classes)
	#> ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
	```
	Should we check these patterns to be consistent with `lintr`?
3. We can change diagnostics and suggestions into actual code.

I can open a PR to help improve in the three areas. What's your idea @etiennebacher ?

PR #555, opened 2026-07-18
feat: Improve `expect_s3_class_linter`

Yousa-Mirage:
## Summary

Part of #8 . Related to #232 .

This PR expands and refactors the `expect_s3_class` rule to cover the additional patterns supported by `lintr::expect_s3_class_linter()`, while preserving the safeguards discussed in #232.

## Changes

- Refactored the rule into separate branches for:
  - `expect_equal()` and `expect_identical()`
  - `expect_true(is.<class>(x))`
  - `expect_true(inherits(x, class))`
- Added support for `expect_true(is.<class>(x))` using the same curated S3 predicate allowlist as [`lintr`](https://github.com/r-lib/lintr/blob/main/R/expect_s3_class_linter.R).
- Added support for `expect_true(inherits(x, class))`.
- Added support for dynamic class expressions such as:
  - `expect_equal(class(x), classes)`
  - `expect_equal(classes, class(x))`
  - `expect_true(inherits(x, classes))`
- Dynamic class expressions now produce a diagnostic but do not receive an automatic fix because their value cannot be determined statically.
- Diagnostics and suggestions now include the actual expressions instead of generic `x` and `k` placeholders.
- Added some new tests and fixed a misspelled test.

## Remain

- Existing `expect_equal(class(x), class)` and `expect_identical(class(x), class)` matches retain their lint ranges and automatic fixes.
- Predicates that do not reliably test an S3 class, such as `is.matrix()` and `is.nan()`, remain ignored as #232 .
- Calls with additional expectation arguments such as `info`, `label`, or `expected.label` remain ignored.
- The rule does not add `exact = TRUE` automatically. Whether exact class equality is intended cannot be inferred statically, so users must add it when appropriate.
- Automatic fixes for statically supported class literals remain safe.
- Namespace and comment-preservation behavior remains unchanged.
- The original tests remains unchanged.

### DIFF
--- a/crates/jarl-core/src/lints/testthat/expect_s3_class/expect_s3_class.rs
+++ b/crates/jarl-core/src/lints/testthat/expect_s3_class/expect_s3_class.rs
@@ -28,7 +28,10 @@ use biome_rowan::{AstNode, AstSeparatedList};
 /// This rule is **disabled by default**. Select it either with the rule name
 /// `"expect_s3_class"` or with the rule group `"TESTTHAT"`.
 ///
-/// This rule has a safe automatic fix but doesn't report cases where:
+/// This rule has a safe automatic fix for statically supported class names.
+/// Dynamic class expressions are reported without an automatic fix.
+///
+/// This rule doesn't report cases where:
 ///
 /// * an `is.*()` predicate is not known to test an S3 class. For example,
 ///   `is.matrix(x)` does not imply that `x` is an S3 object.
@@ -39,12 +42,6 @@ use biome_rowan::{AstNode, AstSeparatedList};
 ///   testthat::expect_s3_class(1L, "integer")
 ///   ```
 ///   For those cases, it is recommended to use `expect_type()` instead.
-///
-/// * the `expected` object could have multiple values, such as:
-///   ```r
-///   testthat::expect_equal(class(x), c("foo", "bar"))
-///   testthat::expect_equal(class(x), vec_of_classes)
-///   ```
 ///
 /// Finally, the intent of the test cannot be inferred with the code only, so
 /// the user will have to add `exact = TRUE` if necessary.
@@ -87,10 +84,8 @@ fn check_expect_class_comparison(
         return Ok(None);
     }
 
-    let object_argument =
-        unwrap_or_return_none!(get_arg_by_name_then_position(&arguments, "object", 1));
-    let expected_argument =
-        unwrap_or_return_none!(get_arg_by_name_then_position(&arguments, "expected", 2));
+    let (object_argument, expected_argument) =
+        unwrap_or_return_none!(get_two_arguments(&arguments, "object", "expected"));
 
     let object_value = unwrap_or_return_none!(object_argument.value());
     let expected_value = unwrap_or_return_none!(expected_argument.value());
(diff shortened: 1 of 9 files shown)


crates/jarl-core/src/lints/testthat/expect_s3_class/expect_s3_class.rs
```diff
@@ -50,101 +51,254 @@ use biome_rowan::{AstNode, AstSeparatedList};
...
 
-    let object_value = unwrap_or_return_none!(object.value());
-    let expected_value = unwrap_or_return_none!(expected.value());
-
-    // Find patterns like `expect_equal(class(x), 'y')` and `expect_equal('y', class(x))`.
-    let (class_arg, other_arg) = if let Some(object_call) = object_value.as_r_call() {
-        let obj_fn = object_call.function()?;
-        let obj_fn_name = get_function_name(obj_fn);
+    let (object_argument, expected_argument) =
+        unwrap_or_return_none!(get_two_arguments(&arguments, "object", "expected"));
```
etiennebacher: Let's not create the helper `get_two_arguments()` just for this rule. I think using `get_arg_by_name_then_position()` would be concise enough.

crates/jarl-core/src/lints/testthat/expect_s3_class/expect_s3_class.rs
```diff
@@ -50,101 +51,254 @@ use biome_rowan::{AstNode, AstSeparatedList};
...
+        can_fix: true,
+    }))
+}
+
+fn as_class_call(expression: &AnyRExpression) -> anyhow::Result<Option<&RCall>> {
+    let Some(call) = expression.as_r_call() else {
+        return Ok(None);
+    };
+
+    Ok((get_function_name(call.function()?) == "class").then_some(call))
```
etiennebacher: I think that's cleaner:

```
if let Some(call) = expression.as_r_call()
        && get_function_name(call.function()?) == "class"
    {
        return Ok(Some(call));
    } else {
        return Ok(None);
    };
```

crates/jarl-core/src/lints/testthat/expect_s3_class/expect_s3_class.rs
```diff
@@ -27,7 +28,13 @@ use biome_rowan::{AstNode, AstSeparatedList};
...
 /// `"expect_s3_class"` or with the rule group `"TESTTHAT"`.
 ///
-/// This rule has a safe automatic fix but doesn't report cases where:
+/// This rule has a safe automatic fix for statically supported class names.
+/// Dynamic class expressions are reported without an automatic fix.
+///
+/// This rule doesn't report cases where:
+///
+/// * an `is.*()` predicate is not known to test an S3 class. For example,
+///   `is.matrix(x)` does not imply that `x` is an S3 object.
```
etiennebacher:
```suggestion
/// * the `is.*()` predicate does not test an S3 class. For example, `is.matrix(x)` does 
///   not imply that `x` is an S3 object.
```

crates/jarl-core/src/lints/testthat/expect_s3_class/expect_s3_class.rs
```diff
@@ -50,101 +51,254 @@ use biome_rowan::{AstNode, AstSeparatedList};
...
+    )))
+}
+
+struct ClassCheck {
+    object_text: String,
+    class_text: String,
+    can_fix: bool,
 }
 
+fn extract_inherits_check(arguments: &RArgumentList) -> anyhow::Result<Option<ClassCheck>> {
```
etiennebacher, line 227: Please add a small comment and example for this and other helper functions.

etiennebacher (changes requested):
Thanks, I still need to review the tests but I already have some comments to adddress

etiennebacher (approved):
Looks good, thanks! I have tested this on a few packages and didn't see new failures because of this fix, except in `parsnip` which is why I ignored a few more `Matrix` classes

PR #556, opened 2026-07-18
feat: Implement `rep_times_ignored` rule

Yousa-Mirage:
Part of #8 .

Initially we wanted to implement `lintr::rep_len_linter`, but after evaluation we decided to implement a new `rep_times_ignored` rule.


crates/jarl-core/src/rule_set.rs
```diff
@@ -559,6 +559,13 @@ declare_rules! {
         fix: Safe,
         min_r_version: None,
     },
+    RepLen => {
+        name: "rep_len",
+        categories: [Read],
+        default: Enabled,
```
etiennebacher, line 565: Since this is just about readability and the fix is unsafe, I would rather have this disabled by default (also needs to be updated in docs).

crates/jarl-core/src/lints/base/rep_len/rep_len.rs
```diff
@@ -0,0 +1,79 @@
...
+
+/// Version added: 0.6.0
+///
+/// ## What it does
+///
+/// Checks for usage of `rep(x, length.out = n)`.
+///
+/// ## Why is this bad?
+///
+/// `rep(x, length.out = n)` calls `rep_len(x, n)` "under the hood". The latter
```
etiennebacher:
```suggestion
/// `rep(x, length.out = n)` calls `rep_len(x, n)` internally. The latter
```

crates/jarl-core/src/lints/base/rep_len/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+    fn snapshot_lint(code: &str) -> String {
+        format_diagnostics(code, "rep_len", None)
+    }
+
+    #[test]
+    fn test_no_lint_rep_len() {
+        expect_no_lint("rep(x, y)", "rep_len", None);
+        expect_no_lint("rep(1:10, 2)", "rep_len", None);
+        expect_no_lint("rep(1:10, 10:1)", "rep_len", None);
+        expect_no_lint("rep(1:10, 10:1, length.out = 50)", "rep_len", None);
```
etiennebacher: This is reported by `lintr` so we should too if possible:

```r
> lintr::lint("rep(1:10, 10:1, length.out = 50)\n", linters = lintr::rep_len_linter())
'[<text>:1:1](vscode-file://vscode-app/usr/share/positron/resources/app/out/vs/code/electron-browser/workbench/workbench.html#)': warning: [rep_len_linter] Use rep_len(x, n) instead of rep(x, length.out = n).
rep(1:10, 10:1, length.out = 50)
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

crates/jarl-core/src/lints/base/rep_len/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+        expect_no_lint("rep(1:10, times = 10:1, length.out = 50)", "rep_len", None);
+        expect_no_lint("rep(1:10, 10:1, 50)", "rep_len", None);
```
etiennebacher: Same here. Please double check the tests with the `lintr` test suite: https://github.com/r-lib/lintr/blob/main/tests/testthat/test-rep_len_linter.R

etiennebacher (changes requested):
Thanks! Seems that we report less cases than `lintr` here, see comments

Yousa-Mirage:
Thank you. I added support for the three patterns that `lintr` supports:

```r
rep(x, times, length.out = n)
rep(x, times = times, length.out = n)
rep(x, times, n)
```

Now the difference from `lintr` is only in some edge cases:

<table>
<tr>
 <td>
 <td>Code
 <td>Jarl
 <td>lintr
<tr>
 <td>1
 <td>rep(x, eac = 4, length.out = 50)
 <td>No
 <td>Yes
<tr>
 <td>2
 <td>rep(x, other = 4, length.out = 50)
 <td>No
 <td>Yes
<tr>
 <td>3
 <td>rep(x, length.out =)
 <td>No
 <td>Yes
<tr>
 <td>4
 <td>rep(x, length.out = NA)
 <td>No
 <td>Yes
<tr>
 <td>5
 <td>rep(x = 1:2, 2, 50)
 <td>Yes
 <td>No
<tr>
 <td>6
 <td>rep(1:2, times = 2, 50)
 <td>Yes
 <td>No
</table>

I think Jarl's current behavior is more reasonable in these circumstances. In R, the expression `rep(x, length.out =)` is actually legal, which is really amazing and confusing.

etiennebacher:
Re-reading the description of the lint and the docs of `?rep`, I'm not convinced we should report all those cases and I don't think this should be a readability rule. For example, it's not clear to me that `rep_len(1:100, 5)` is more readable than `rep(1:100, length.out = 5)`.

This rule could be adapted to only report cases where the user passes both `times` and `length.out` since this is likely a mistake (`?rep` says "If both are given, `length.out` takes priority and `times` is ignored."), so it could be in the category "Suspicious". This should cover cases where args are named or unnamed, e.g. `rep(1:10, 10:1, 50)` should be reported.

What do you think?

Edit: this means the rule should be renamed, maybe `rep_ignored_arg`? (not a big fan of this name, open to alternatives)

Yousa-Mirage:
I agree with you on these points. `rep_len()` is really no better than `rep(x, length.out = n)`. For this rule, we can just check for problematic situations when the `rep()` function is used. I will make some modifications.

Yousa-Mirage:
What do you think of `rep_times_ignored`?

etiennebacher:
Sounds good to me, curious to see the ecosystem checks for this one.

Yousa-Mirage:
OK, I've finished the refactor work. Now this linter just checks if both `times` and `length.out` appear at the same time.

etiennebacher (approved):
Thanks for addressing my comments, that was a big change to make. I've just enabled this rule by default since it is now in the "suspicious" category, so I think it is important to report it to users. The low number of violations in ecosystem checks also means it shouldn't appear too often anyway.

PR #560, opened 2026-07-20
feat: Improve `which_grepl_linter`

Yousa-Mirage:
This PR improves the which_grepl rule by:

- Detecting `grepl()` passed through the named `x` argument of `which()`, such as `which(x = grepl(...))`.
- Reporting calls where `which()` has additional arguments such as `arr.ind` or `useNames`, without automatic fix.
- Supporting three-stage native pipes like `values |> grepl(pattern = "a") |> which()` which is safely fixed to `grep(values, pattern = "a")`.
- Preserving all `grepl()` arguments in diagnostics and fixes.
- Skipping invalid or ambiguous pipe calls such as `x |> grepl() |> which()`.
- Showing the actual source and replacement expressions in diagnostic messages instead of generic `pattern` and `x` placeholders.
- Simplifying the rule documentation example.
- Added more tests.


crates/jarl-core/src/lints/base/which_grepl/mod.rs
```diff
@@ -31,9 +31,35 @@ mod tests {
...
+          |
+        1 | which(x = grepl('^a', x))
+          | ------------------------- `which(x = grepl('^a', x))` is less efficient than `grep('^a', x)`.
+          |
+          = help: Use `grep('^a', x)` instead.
+        Found 1 error.
+        "
+        );
+        assert_snapshot!(
+            snapshot_lint("which(grepl('^a', x), arr.ind = TRUE)"),
```
etiennebacher, line 54: Do we want to report cases where `arr.ind = TRUE`? 

`?which` says:

> If arr.ind == TRUE and x is an array (has a dim attribute), the result is arrayInd(which(x), dim(x), dimnames(x)), namely a matrix whose rows each are the indices of one element of x; see Examples below.

so it seems to me `grepv()` wouldn't be equivalent here.

crates/jarl-core/src/lints/base/which_grepl/mod.rs
```diff
@@ -58,27 +87,49 @@ mod tests {
...
             @"
         warning: which_grepl
          --> <test>:1:1
           |
-        1 | / grepl('^a', x) |> 
+        1 | / grepl('^a', x) |>
         2 | |  which()
-          | |________- `which(grepl(pattern, x))` is less efficient than `grep(pattern, x)`.
+          | |________- `grepl('^a', x) |>
+         which()` is less efficient than `grep('^a', x)`.
```
etiennebacher: I think I'd rather keep the generic message here rather than replacing it with the code that is reported since it breaks a line in the output.

crates/jarl-core/src/lints/base/which_grepl/which_grepl.rs
```diff
@@ -18,47 +20,86 @@ pub struct WhichGrepl;
 /// ## Example
 ///
 /// ```r
-/// x <- c("hello", "there")
-/// which(grepl("hell", x))
```
etiennebacher, line 22: Why is this removed? The `x` is needed to be able to run the code if someone wants to copy-paste it, and the `which()` would output one value (contrary to the second `which()` that return `integer(0).`

crates/jarl-core/src/lints/base/which_grepl/which_grepl.rs
```diff
@@ -18,47 +20,86 @@ pub struct WhichGrepl;
...
 /// which(grepl("foo", x))
 /// ```
 ///
 /// Use instead:
 /// ```r
-/// x <- c("hello", "there")
-/// grep("hell", x)
```
etiennebacher, line 29: Same

crates/jarl-core/src/lints/base/which_grepl/which_grepl.rs
etiennebacher, line 21: Can you mention here that this rule sometimes has an automatic fix?

etiennebacher (changes requested):
Thanks!

crates/jarl-core/src/lints/base/which_grepl/which_grepl.rs
Yousa-Mirage: Oops, I was trying to keep just one example but forgot to replace `x`. I modified this manually and I wasn't careful enough.

crates/jarl-core/src/lints/base/which_grepl/mod.rs
Yousa-Mirage: `grepl()` drops the dimensions of its input and returns a plain logical vector. Therefore, in this specific pattern, `which(grepl(...), arr.ind = TRUE)` still returns linear indices and is equivalent to `grep(...)`.

``` r
x <- matrix(c("a", "b", "a", "c"), nrow = 2)
x
#>      [,1] [,2]
#> [1,] "a"  "a" 
#> [2,] "b"  "c"

x == "a"
#>       [,1]  [,2]
#> [1,]  TRUE  TRUE
#> [2,] FALSE FALSE
which(x == "a", arr.ind = TRUE)
#>      row col
#> [1,]   1   1
#> [2,]   1   2

grepl("a", x)
#> [1]  TRUE FALSE  TRUE FALSE
which(grepl("a", x), arr.ind = TRUE)
#> [1] 1 3

grep("a", x)
#> [1] 1 3
```

So, I think it is still useful to report it, and when `arr.ind` is a logical literal, we could fix it safely.

etiennebacher: Yeah I got confused by the fact that `grepl()` will never return an array here. But then can't we just apply the autofix no matter the value of `arr.ind`, i.e. fix `which(grepl("a", x), arr.ind = foo)`?

crates/jarl-core/src/lints/base/which_grepl/which_grepl.rs
```diff
@@ -15,6 +17,17 @@ pub struct WhichGrepl;
...
+/// This rule has an automatic fix for direct calls where `which()` only
+/// contains the `grepl()` call, and for pipe chains where the final `which()`
+/// has no arguments and the piped value can be unambiguously assigned to
+/// `grepl()`'s `pattern` or `x` argument.
+///
+/// Calls with additional arguments to `which()` are reported but not fixed
+/// because those arguments cannot generally be preserved by replacing
+/// `which()` with `grep()`. The exception is a literal `arr.ind = TRUE` or
+/// `FALSE`: `grepl()` returns a vector without dimensions, so `arr.ind` has no
+/// effect.
```
etiennebacher: I think this paragraph can be removed since `arr.ind` is for arrays only (discussed before) and it is unclear what `useNames` is for because the docs explain what it does in `arrayInd()` only.

etiennebacher (changes requested):
There's just a comment above that needs to be resolved about whether we should fix all cases that have `arr.ind` (and then whether docs should be updated accordingly) and then this should be good to go.

I made a small change to show a generic message instead of a very precise replacement in the `suggestion` field. This is because one of my previous comment was about having a generic message in the `message` field as well and it would be weird to show a precise message for the suggestion only IMO. Also, I'm a bit cautious about using user's code in the diagnostic because this is what appears in the IDE when the violation is hovered, and this could lead to very long messages.

crates/jarl-core/src/lints/base/which_grepl/mod.rs
Yousa-Mirage: My main concern is that this fix might change the behavior if `foo()` in `arr.ind = foo` has some kind of side effect or reports an error (such as `arr.ind = stop("error")`). But this may be very rare. If that's feasible for you, we can ignore the `arr.ind` parameter entirely.

crates/jarl-core/src/lints/base/which_grepl/which_grepl.rs
Yousa-Mirage: It seems that we can ignore all parameters of which as long as we decide not to care about dynamic expressions or functions that may cause side effects. But I think very few people would write like this:
```r
which(
  grepl("a", x),
  arr.ind = {
    message("evaluated")
    TRUE
  }
)
```

Yousa-Mirage:
I now completely ignore the `arr.ind` and `useNames` parameters. Whatever their values ​​are will be discarded in the fix. This should not break normal code, since normally these two parameters have no effect on the results produced by `grepl`.

crates/jarl-core/src/lints/base/which_grepl/which_grepl.rs
etiennebacher: Yeah I think so too

etiennebacher (approved):
Thanks for addressing my comments!

PR #572, opened 2026-07-29
feat: Implement `list_comparison_linter`

Yousa-Mirage:
Part of #8 . The rule in `lintr` is https://lintr.r-lib.org/reference/list_comparison_linter.html. 
Currently this rule is enabled by default and has no auto-fix, because the data type in `list` cannot be inferred.


etiennebacher:
I haven't forgotten about this, it's just low priority to me since there are no occurrences in ecosystem checks and I also don't have any violations in any of my local projects.

Yousa-Mirage:
No matter at all. You can handle it anytime as you like. This is not an important rule.

PR #573, opened 2026-07-29
feat: Implement `nonportable_path_linter`

Yousa-Mirage:
Part of #8 . This rule is from https://lintr.r-lib.org/reference/nonportable_path_linter.html .

This rule is disabled by default because it can produce many false positives for strings containing "/" such as regular expressions. The problem of false positives also exists in `lintr`. If you feel it is inappropriate, we do not need to implement this linter.

### DIFF
--- a/crates/jarl-core/src/analyze/anyvalue.rs
+++ b/crates/jarl-core/src/analyze/anyvalue.rs
@@ -2,10 +2,14 @@ use crate::checker::Checker;
 use crate::rule_set::Rule;
 use air_r_syntax::AnyRValue;
 
+use crate::lints::base::nonportable_path::nonportable_path::nonportable_path;
 use crate::lints::base::numeric_leading_zero::numeric_leading_zero::numeric_leading_zero;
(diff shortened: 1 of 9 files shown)


etiennebacher:
I can't take a detailed look at this right now, but I suspect this is going to raise many, many false positives (in fact, the workflow to post results of the ecosystem checks failed because the PR comment would contain too much text, which suggests that there are many violations in all repos). 

I'll take a look later but I don't think Jarl has the required capabilities for this rule yet (e.g. to know whether an object is used in a call that reads a file).

Yousa-Mirage:
Yes. If you feel this rule is inappropriate after evaluation, we may not implement this rule. Because we can't differentiate between normal strings with slashes and file paths, for that I can't thought of anything that works. Maybe this rule itself is not a good rule. I don't think adding `file.path` to every file path string is a good solution. If you feel bad, please feel free to close this PR.

etiennebacher:
Thank you for the effort but I don't think this is worth implementing for now.

PR #579, opened 2026-08-03
fix: Ensure that `--fix-only` doesn't apply unsafe fixes

Yousa-Mirage:
The behavior of `--fix-only` is incorrect. When only `jarl check . --fix-only` is used, the fix isn't actually applied. However, when use `jarl check . --fix --fix-only`, unsafe fixes are applied accidentally. For example:

```r
!all.equal(x, y)  # Have an unsafe fix
!(x > y)  # Have a safe fix
```

After `jarl check . --fix-only`, jarl just reports lints without auto-fix. But after `jarl check . --fix --fix-only`, both are fixed:

```r
!isTRUE(all.equal(x, y))  # Have an unsafe fix
x <= y  # Have a safe fix
```

This PR fixed this issue by using `check_config.fix || check_config.fix_only`.

### DIFF
--- a/crates/jarl-core/src/config.rs
+++ b/crates/jarl-core/src/config.rs
@@ -64,7 +64,7 @@ pub struct Config {
     /// field `rules`. If we apply fixes too, then this might be different from
     /// `rules` because it may filter out rules that have unsafe fixes.
     pub rules_to_apply: RuleSet,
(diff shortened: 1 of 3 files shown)


etiennebacher (approved):
Perfect, thanks!

PR #581, opened 2026-08-03
fix: Traverse subset and extraction objects

### LINKED ISSUE #580: [BUG] Lint violations are missed on the left-hand side of `[`, `[[`, `$`, and `@`
Just see this image:

Image

In `RSubset` and `RSubset2`, the `children.function()` part (`x` of `x[...]`) is not checked recursively. Same for `children.left()` of `RExtractExpression`.

I can open a PR to fix this.

Yousa-Mirage:
Fix #580 . Check `children.function()` recursively.


Yousa-Mirage:
It looks like rules like `true_false_symbol` are producing a lot of false positives. But I think this should be fixed in the `true_false_symbol` rules, e.g. ignoring when `T` was found to be to the left of `RSubset` or `RSubset2`, thus avoiding false positives like `T[1:2]`.

crates/jarl/tests/integration/edge_cases.rs
```diff
@@ -112,3 +112,71 @@ fn test_jarl_with_tabs() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+#[test]
+fn test_jarl_traverses_subset_and_extraction_objects() -> anyhow::Result<()> {
```
etiennebacher: I don't think this should be in `edge_cases.rs`, it was clearly an oversight from my side. Can you add those extra cases in the tests of `any_is_na` instead?

etiennebacher (changes requested):
Thank you, that was a big mistake. 

I didn't check all the ecosystem results but the deleted violations in `dplyr` look like false negatives to me. Could you verify?

> It looks like rules like true_false_symbol are producing a lot of false positives. But I think this should be fixed in the true_false_symbol rules, e.g. ignoring when T was found to be to the left of RSubset or RSubset2, thus avoiding false positives like T[1:2].

Agreed, can you either open an issue or make a new PR on top of this one with this fix?

Yousa-Mirage:
The deleted violations in `dplyr` are all code in comments. I don't know why they are reported before.

etiennebacher:
Jarl reports violations in `@examples` section (https://jarl.etiennebacher.com/reference/config-file.html#check-roxygen) but can you check whether this PR also removes these violations when they are not in `#'` comments?

Yousa-Mirage:
I can't reproduce the 8 removed violations in the check for `dplyr`.

I don't think this PR will affect those cases. Maybe CI has some mistake? I don't know.

etiennebacher:
I think the issue is #583: before this PR, we didn't parse the LHS of `[[` so this didn't lead to parsing errors. After this PR, we parse it, leading to a parsing error for the entire file and therefore removing the 8 existing violations.

Yousa-Mirage:
Very interesting 😮

Yousa-Mirage:
> I didn't check all the ecosystem results but the deleted violations in `dplyr` look like false negatives to me.

Fine. This issue has been fixed.

PR #582, opened 2026-08-03
fix: Ignore subset objects in `true_false_symbol`

Yousa-Mirage:
Depends on #581, part of #580 

- Avoid false positives when `T` or `F` is the object of `[` or `[[`, such as `T[1:2]`.
- Continue reporting `T` and `F` when used as subset indices or as logical values inside the subset object.
- Add regression tests covering both reported and allowed cases.

Verification

  ```r
T[1:2]  # no lint
T[[1]]  # no lint
x[T]    # lint
x[[F]]  # lint
T$F    # no lint
```


etiennebacher:
FYI, if you press the "Edit title" icon at the top, you should be able to change the reference branch for this PR to put #581 instead. (Not sure whether this is possible when you're contributing via a fork)

Yousa-Mirage:
> FYI, if you press the "Edit title" icon at the top, you should be able to change the reference branch for this PR to put #581 instead. (Not sure whether this is possible when you're contributing via a fork)

No, I can't. But I think once #581 is merged, we can just merge from main?

And I noticed that the removed violations of `dplyr` occured here too. I really don't know why. But I will continue to check these new violations to try to avoid false positives.

Yousa-Mirage:
I took a look and found that after fixing `true_false_symbol`, these new ecological diagnostics have few false positives. There are indeed some issues that go undetected because they are on the left side of `$` or `[`. Three false positives are:

- The trailing comma in `tribble()` triggers `missing_argument` in https://github.com/pola-rs/r-polars.
- https://github.com/wch/r-source/tree/b5cf23a805f4305852f55c4148f2a302b06844b5/src/library/stats/R/mlm.R#L329 uses `T` as matrix var in `%*%`.
- https://github.com/TheAlgorithms/R/tree/f65344fabeea0ae33c1ca321af709447c1a8a4c2/quantitative_finance/monte_carlo_simulation.r#L22 uses `T` as time parameter.

PR #592, opened 2026-08-03
feat: Treat `do.call()`, `on.exit()` and loop assignments as uses in `unused_object`

etiennebacher:


etiennebacher (commented):
BLOCKED: `survey` results look wrong

Issue #605, opened 2026-08-10
New rule: flag assignment on if


christopherkenny:
One of the technically valid, but often undesirable, things that you can do in R is assign off of an if statement. For example:

```r
df <- if (cond) {
  data.frame()
}
```
which makes `df` into `NULL` if `!cond`.

There are other variations, such as:

```r
df <- if (cond) {
  data.frame()
} else {
  mtcars
}
```
or inline in functions:

```
some_fn(arg = if (cond) 1 else sqrt(pi))
```

I think it would be great to have a rule (opt-in) which yells about this. I've noticed the newest batch of GPT-5.6 models, especially Sol, seem to really like this pattern.

Some of the peerage of the tidyverse have had an interesting conversation (https://github.com/tidyverse/design/issues/71) and a shorter discussion to ban it (https://github.com/tidyverse/design/issues/131). I cite to these to say "it's not just me who might use it" and "non-default is probably the right class for it".

[Proof of concept](https://github.com/christopherkenny/jarl/tree/assignment-on-if) to show it's feasible.

etiennebacher:
Thanks, there are several cases here.

### No else branch

This is indeed likely a mistake because the previous value of the assigned object isn't kept if the condition is false:

``` r
df <- 1
df <- if (FALSE) {
  data.frame()
}
df
#> NULL
```

Note that this should also capture else-if without else:

``` r
df <- 1
df <- if (FALSE) {
  data.frame()
} else if (FALSE) {
  2
}
df
#> NULL
```

This could be a "Suspicious" rule (opt-out).

### Assignment outside branches

```r
df <- if (cond) {
  data.frame()
} else {
  mtcars
}
```
There was actually a long discussion on rOpenSci's Slack about whether people accept that or not. I think we could make this a readability rule (opt-in) only if the branches consist of several statements because after reading the branch it can be hard to remember that the result is assigned to an object:

```r
df <- if (cond) {
  x <- mtcars[1:3, ]
  x$drat <- x$drat * 2
  x
} else {
  x <- mtcars[1:5, ]
  x$drat <- x$drat * 10
  x
}
```

### Inline if-else in function args

Agreed this could be a rule, although probably a "Readability" one only (opt-in).


### How many rules?

Although those are all related to `if` conditions, I'd rather keep some granularity and have one rule per case so that people can choose which to select or exclude. 

Happy to review separate PRs for those.

---

[...]

PR #606, opened 2026-08-10
feat: Add `assignment_on_if_no_else` rule

christopherkenny:
Part 1 / 3 of #605

This PR implements the suspicious branch of #605 for the case of assignment off of an `if` or subsequent `if else` lacking a final `else`.

Happy to adjust the name, but given the other rule names, this seems like the right level of specificity.

In line with the package policy, I want to disclose that the code was primarily generated by GPT-5.6 Luna Max, with several rounds of review by me. My Rust is not nearly as good as my R, but everything looks in order.


etiennebacher:
Looking at the ecosystem results, it seems to me that most of the time developers are aware that the object will be `NULL` if the `if` condition is false. I think we should focus on reporting cases where the object is defined and never used until another assignment based on an incomplete `if` statement.

For example, this should be reported:

```r
x <- 1
x <- if (foo) y
```
because it is likely a bug where the user thought `x <- 1` would be used if `foo` is false, which isn't the case. 

However, this shouldn't be reported:

```r
x <- if (foo) y
```
because the user might be well aware that `x` will be `NULL` if `foo` is false.

This adds some complexity to the rule because now we need to know if the target of the assignment has been defined but not used before (and the scope of the previous assignment must be reachable by the assignment using `if`) but I think it could be doable now that there is `jarl-semantic`. I don't know that for sure however, this would need some exploration.

etiennebacher:
@christopherkenny just checking in, do you want to try to address my previous comment? No pressure from me if you want to but just lack some time, but it's not trivial to add semantic analysis here so I can also finish this PR if you prefer.

christopherkenny:
Thanks for the review. I've taken a stab at it. In the example script below, we see the expected reported condition.

```r
condition <- TRUE

# Reported: the incomplete `if` can overwrite the earlier, unused value.
reused <- 1
reused <- if (condition) 2

# Not reported: there is no earlier value to overwrite.
standalone <- if (condition) 2

# Not reported: the earlier value is used before it is overwritten.
used <- 1
print(used)
used <- if (condition) 2

# Not reported: the earlier assignment is in a mutually exclusive branch.
if (condition) {
  branch <- 1
} else {
  branch <- if (condition) 2
}
```

returns:

```
warning: assignment_on_if_no_else
  --> scripts\assignment_on_if_no_else_demo.R:5:1
   |
5 | reused <- if (condition) 2
   | -------------------------- This assignment can overwrite the previous value with `NULL` when no `if` branch is taken.
   |
   = help: Move the assignment into the `if` branches or add a final `else` branch.
```

crates/jarl-core/src/analyze/document.rs
```diff
@@ -31,6 +33,22 @@ pub(crate) fn check_document(
...
+            let root = first.ancestors().last().unwrap_or_else(|| first.clone());
+            Some(SemanticInfo::build(
+                &root,
+                &expressions,
+                semantic,
+                &checker.source_index_cache,
+                &checker.loaded_packages,
+                &unevaluated_ranges,
+            ))
+        });
```
etiennebacher, line 51: I think this can be moved within `if checker.is_rule_enabled(Rule::AssignmentOnIfNoElse)` below (to avoid conditioning twice on this rule).

crates/jarl-core/src/lints/base/assignment_on_if_no_else/assignment_on_if_no_else.rs
```diff
@@ -0,0 +1,125 @@
+use crate::diagnostic::*;
+use crate::rule_set::Rule;
+use air_r_syntax::*;
+use biome_rowan::AstNode;
+use jarl_semantic::{SemanticInfo, assignment_lhs_name};
+
+pub struct AssignmentOnIfNoElse;
+
+/// Version added: 0.6.0
```
etiennebacher, line 9: 0.7.0

crates/jarl-core/src/lints/base/assignment_on_if_no_else/assignment_on_if_no_else.rs
```diff
@@ -0,0 +1,125 @@
...
+///   data.frame()
+/// }
+///
+/// value <- 1
+/// value <- if (a) {
+///   1
+/// } else if (b) {
+///   2
+/// }
+/// ```
```
etiennebacher, line 37:
```suggestion
/// ```r
/// df <- mtcars
/// # If `condition` is `FALSE` then `df` becomes `NULL` instead of 
/// # keeping its original value.
/// df <- if (condition) {
///   mtcars[1:5,]
/// }
///
/// value <- 1
/// # If neither `a` nor `b` are `TRUE` then `value` becomes `NULL`
/// instead of keeping its original value.
/// value <- if (a) {
///   1
/// } else if (b) {
///   2
/// }
/// ```
```

crates/jarl-core/src/toml.rs
```diff
@@ -36,7 +36,7 @@ use crate::settings::Settings;
 #[derive(Debug)]
 pub enum ParseTomlError {
     Read(PathBuf, io::Error),
-    Deserialize(PathBuf, toml::de::Error),
+    Deserialize(PathBuf, Box<toml::de::Error>),
```
etiennebacher, line 39: I don't think changes in this file are necessary

crates/jarl-core/src/lints/base/assignment_on_if_no_else/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    #[test]
+    fn test_assignment_on_if_no_else_is_suspicious_and_enabled_by_default() {
+        assert!(Rule::AssignmentOnIfNoElse.is_enabled_by_default());
+        assert!(Rule::AssignmentOnIfNoElse.has_category(Category::Susp));
+    }
+
+    #[test]
+    fn test_lint_assignment_on_if_no_else() {
```
etiennebacher, line 20: This should test the following cases as well:

1. should report

```r
df <- 1
foo <- function() {
  df <- if (cond) {
    data.frame()
  }
  df
}
```

2. should report

```r
bar <- function() {
  df <- 1
  foo <- function() {
    df <- if (cond) {
      data.frame()
    }
    df
  }
}
```

3. should not report

```r
foo <- function() {
  df <- 1
  df
}

df <- if (cond) {
  data.frame()
}
```

etiennebacher (changes requested):
Thanks, this looks good but I think the semantic analysis should also do more to detect similar cases across scopes, see the tests suggested below.

Eventually it would be great that the message shows where the object in question was defined before, but this requires more changes to Jarl and doesn't have to be addressed in this PR.

Issue #565, opened 2026-07-21
Jarl 0.6.0


etiennebacher:
- [x] Fix #608
- [x] Review new rules enabled by default: how many new violations does that make?  
   - actually pretty annoying to do since I'd have to modify the ecosystem checks, but I have reviewed the list of new rules and whether they are enabled by default  
- [x] Fix #576 
- [x] Finish blog post
- [x] Finish branch to reduce binary size (#633)

PR #633, opened 2026-08-21
build: Reduce binary size

etiennebacher:
Part of #565


etiennebacher:
```
hyperfine --warmup 5 '/home/etienne/Desktop/Git/jarl/target/dist/jarl check . --statistics' '/home/etienne/Desktop/Git/jarl/target/release/jarl check . --statistics' --ignore-failure --runs 20
```
on `r-source` gives very different results at each run and it's not clear at all which of `dist` or `release` is faster, which is good news.

PR #634, opened 2026-08-21
[DO NOT MERGE] Run benchmarks

etiennebacher:
Benchmarks couldn't run in #633 because the `benchmark` profile didn't exist on the main branch yet. Results here should be compared to other PRs.


etiennebacher:
Got mixed up, benchmarks don't run with `dist` profile anyway

PR #657, opened 2026-08-23
feat: Implement `undesirable_operator` rule

christopherkenny:
Part of #8. The rule in `lintr` is [`undesirable_operator_linter`](https://lintr.r-lib.org/reference/undesirable_operator_linter.html). I started with the special backtick option that `lintr` has, but that seemed to offer unnecessary complexity.

This PR adds support for the `undesirable_operator` here, closely mirroring the implementation of `undesirable_function`. I set it as "Correctness" and on by default, to match the function version. There is no auto-fix option.

One thing that I want to flag ahead is that this creates duplicate behavior for `:::`, since there's a special case for that already. It would be simple to drop that and update tests if you don't want that included.

No rush here, just a feature I would like to use when you have time to look.

### DIFF
--- a/crates/jarl-core/src/analyze/expression.rs
+++ b/crates/jarl-core/src/analyze/expression.rs
@@ -52,6 +52,7 @@ pub(crate) fn check_expression(
             }
         }
         AnyRExpression::RExtractExpression(children) => {
+            analyze::extract_expression::extract_expression(children, checker)?;
             check_expression(&children.left()?, checker)?;
         }
         AnyRExpression::RForStatement(children) => {
--- a/crates/jarl-core/src/analyze/extract_expression.rs
+++ b/crates/jarl-core/src/analyze/extract_expression.rs
@@ -1,0 +1,17 @@
+use crate::checker::Checker;
+use crate::lints::base::undesirable_operator::undesirable_operator::undesirable_operator_extract;
+use crate::rule_set::Rule;
+use air_r_syntax::RExtractExpression;
+
+pub fn extract_expression(
+    r_expr: &RExtractExpression,
+    checker: &mut Checker,
+) -> anyhow::Result<()> {
+    if checker.is_rule_enabled(Rule::UndesirableOperator) {
+        checker.report_diagnostic(undesirable_operator_extract(
+            r_expr,
+            &checker.rule_options.undesirable_operator,
+        )?);
+    }
+    Ok(())
+}
--- a/crates/jarl-core/src/lints/base/undesirable_operator/undesirable_operator.rs
+++ b/crates/jarl-core/src/lints/base/undesirable_operator/undesirable_operator.rs
@@ -32,17 +32,14 @@ pub struct UndesirableOperator {
 /// operators = ["$", "@"]
 /// ```
 ///
-/// To add to the defaults or skip backtick-quoted calls to operators:
+/// To add to the defaults:
 ///
 /// ```toml
 /// [lint.undesirable_operator]
 /// extend-operators = ["$", "%in%"]
-/// call-is-undesirable = false
 /// ```
 ///
-/// Specifying both `operators` and `extend-operators` is an error. Setting
-/// `call-is-undesirable = false` still reports the banned operators when
-/// they are used in ordinary expressions.
+/// Specifying both `operators` and `extend-operators` is an error.
 ///
 /// ## Example
 ///
@@ -105,10 +102,6 @@ pub fn undesirable_operator_call(
     ast: &RCall,
     options: &ResolvedUndesirableOperatorOptions,
 ) -> anyhow::Result<Option<Diagnostic>> {
-    if !options.call_is_undesirable {
-        return Ok(None);
-    }
-
     let function = ast.function()?;
     let function_text = function.syntax().text_trimmed().to_string();
     let Some(operator_text) = function_text
--- a/crates/jarl-core/src/lints/base/undesirable_operator/undesirable_operator.rs
+++ b/crates/jarl-core/src/lints/base/undesirable_operator/undesirable_operator.rs
@@ -1,0 +1,129 @@
+use crate::diagnostic::*;
+use crate::lints::base::undesirable_operator::options::ResolvedUndesirableOperatorOptions;
+use crate::rule_set::Rule;
+use air_r_syntax::*;
+use biome_rowan::{AstNode, TextRange};
+
+pub struct UndesirableOperator {
+    pub operator: String,
+}
+
+/// Version added: 0.6.0
+///
+/// ## What it does
+///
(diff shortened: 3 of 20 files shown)


crates/jarl-core/src/analyze/expression.rs
```diff
@@ -52,6 +52,7 @@ pub(crate) fn check_expression(
             }
         }
         AnyRExpression::RExtractExpression(children) => {
+            analyze::extract_expression::extract_expression(children, checker)?;
```
etiennebacher, line 55: I don't think this is needed anymore once this branch is synced with `main`.

crates/jarl-core/src/analyze/extract_expression.rs
```diff
@@ -0,0 +1,17 @@
...
+    checker: &mut Checker,
+) -> anyhow::Result<()> {
+    if checker.is_rule_enabled(Rule::UndesirableOperator) {
+        checker.report_diagnostic(undesirable_operator_extract(
+            r_expr,
+            &checker.rule_options.undesirable_operator,
+        )?);
+    }
+    Ok(())
+}
```
etiennebacher, line 17: Same for this file.

crates/jarl-core/src/lints/base/undesirable_operator/undesirable_operator.rs
```diff
@@ -0,0 +1,122 @@
...
+use crate::lints::base::undesirable_operator::options::ResolvedUndesirableOperatorOptions;
+use crate::rule_set::Rule;
+use air_r_syntax::*;
+use biome_rowan::{AstNode, TextRange};
+
+pub struct UndesirableOperator {
+    pub operator: String,
+}
+
+/// Version added: 0.6.0
```
etiennebacher, line 11: 0.7.0

crates/jarl-core/src/lints/base/undesirable_operator/undesirable_operator.rs
```diff
@@ -0,0 +1,122 @@
...
+///
+/// ## What it does
+///
+/// Checks for use of banned operators.
+///
+/// ## Why is this bad?
+///
+/// Some operators should not appear in production code. For example, `:::`
+/// accesses a package's internal functions, and `<<-` and `->>` assign outside
+/// the current environment.
```
etiennebacher, line 21:
```suggestion
/// Some operators may be undesirable because they make the code less robust,
/// more complex, or more prone to footguns. For example, `:::` accesses a package's
/// internal functions, meaning that they may disappear or behave differently without
/// notice if the package changes. `<<-` and `->>` assign outside the current environment
/// and may lead to code that is harder to predict.
```

This tones down the "should not appear" and explains why they might be undesirable. It might not be the best wording, but you get the idea.

crates/jarl/tests/integration/toml_rule_args.rs
```diff
@@ -598,6 +598,152 @@ pipe = "foo"
     Ok(())
 }
 
+// undesirable_operator ----------------------------------------
+
+#[test]
+fn test_undesirable_operator_options_are_applied() -> anyhow::Result<()> {
```
etiennebacher, line 604: Not needed here, this already tested in the rule tests. Integration tests here should only ensure that we throw nice error messages if 1) both options are passed, or 2) an unknown field is passed.

crates/jarl-core/src/rule_set.rs
```diff
@@ -651,6 +651,13 @@ declare_rules! {
         fix: None,
         min_r_version: None,
     },
+    UndesirableOperator => {
```
etiennebacher, line 654: This new rule supersedes the rule `InternalFunction`, which can be deprecated. See how this is done for the `Browser` rule in this file.

etiennebacher (changes requested):
Thank you and sorry for the delay to review this! This looks good overall, there are a few comments to address but you should first sync this branch with `main`.

crates/jarl-core/src/rule_set.rs
christopherkenny: Deprecated ✔️

crates/jarl/tests/integration/toml_rule_args.rs
christopherkenny: Removed.

crates/jarl-core/src/lints/base/undesirable_operator/undesirable_operator.rs
christopherkenny: Adopted this language, thanks.

crates/jarl-core/src/lints/base/undesirable_operator/undesirable_operator.rs
christopherkenny: Updated to 0.7.0

crates/jarl-core/src/analyze/expression.rs
christopherkenny: Dropped.

crates/jarl-core/src/analyze/extract_expression.rs
christopherkenny: Removed this.

christopherkenny:
Thanks for the review. This should be ready for a second go.

crates/jarl-core/src/lints/base/undesirable_operator/undesirable_operator.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    if !options.operators.contains(operator_text) {
+        return Ok(None);
+    }
+
+    Ok(undesirable_operator(
+        operator_text,
+        function.syntax().text_trimmed_range(),
+    ))
+}
```
etiennebacher, line 124: Do you have an example where this is used? It seems to me anything that hits this should be in `undesirable_function` instead but maybe I'm missing something.

etiennebacher (changes requested):
Thanks!

Actually, `analyze::extract_expression::extract_expression(children, checker)?;` and its related logic should have stayed, it was my mistake about what this line was doing. Sorry! 

The rest looks good, with one other comment to address below.

crates/jarl-core/src/lints/base/undesirable_operator/undesirable_operator.rs
christopherkenny: Suppose that you wanted to ban addition through `+`, then this would catch things like 

```
`+`(1, 2)
```

Maybe it's a weird special case though.

etiennebacher:
Thanks, this looks great. I can't push changes directly for some reason so I'll merge this and do a follow-up PR so that you don't need to go through an extra round of review.

christopherkenny:
Thank you!

Issue #247, opened 2025-12-17
Check naming conventions


AlbertRapp:
Hi Etienne 👋 

Thank you for this very cool project. I really like the extra tooling us R developers are getting lately. Is there a plan to add rules for naming conventions? So that it checks for, say, object names being in snake case like [`lintr::object_name_linter()`](https://lintr.r-lib.org/dev/reference/object_name_linter.html?q=snake#ref-usage) does?

Best,
Albert

pbreheny:
Just chiming in to say:

- I also love Jarl — thank you for developing it
- I also find the object_name_linter helpful and would be interested in seeing it in Jarl, if possible, although I recognize that it's thorny
- I haven't personally experienced false positives; I see false negatives more often.

For example, if you name a variable `predict.args`, lintr won't flag it because it that could be an S3 method for an object with class `args`. In my experience, it's surprisingly good at this and doesn't get fooled even by what I would think are tricky cases like defining novel generics in a package (maybe it looks at NAMESPACE, I'm not sure).

It also won't flag things like `data.frame(X = 5)` or `x$X.Y <- 5` because those are components of objects, not objects themselves.

I haven't tried it with R6 objects, so I'm not sure there.

The closest thing to a false positive that I can think of would be something like this:

```r
data(Owls, package = "glmmTMB")
Owls$any_calls <- Owls$SiblingNegotiation > 1
```

If you don't modify the object, lintr won't say anything. But if you do, lintr notes it (maybe because you could rename it at this point if you chose).

etiennebacher:
Thanks for the extra info and the kind words. I'm hopeful this will be possible in the future once I have a clearer infrastructure to handle S3 methods (and other OOP systems in R), but for now this is not on the roadmap. 

Just in case someone is thinking of implementing it, this issue is not good for first time contributors.

PR #660, opened 2026-08-26
feat: Implement `object_name` linter

christopherkenny:
Another component of #8.

This provides features for assignment-based variable naming conventions. It offers several base cases, similar to other linters, and then also allows users to provide regex for acceptable names.

### DIFF
--- a/artifacts/jarl.schema.json
+++ b/artifacts/jarl.schema.json
@@ -291,7 +291,7 @@
         },
         "object_name": {
           "title": "Options for the `object_name` rule",
-          "description": "Use `styles` to select built-in naming styles and `regexes` to add named\nregular expressions.",
+          "description": "Use `styles` to select built-in naming styles, `regexes` to add named\nregular expressions, and `special-names` / `extend-special-names` to\nconfigure names that are exempt from style checks.",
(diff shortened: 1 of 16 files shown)


etiennebacher:
Thank you, but I think this requires more planning before implementation, cf #247. Checking the object name isn't enough because of S3 methods, R6 classes, and other cases.

Edit: have you tried running this locally? The ecosystem checks are not printed here because there would be too many violations and this is likely due to a lot of false positives.

christopherkenny:
Fair enough on waiting to implement. I tried to keep this to the well-defined scope for basic objects and assignments, but it is weird to leave out some of the other pieces. (edit: more than weird, as this is a source of false positives for some of the CI packages.)

I've run locally, but not on the full CI check. It's possible my temporary test fixtures were "too clean" to catch weird things.

It does make some sense that the defaults here would flag massive numbers of things though for some projects, without expanding the set of acceptable styles for some projects. For example, `ggplot2` doesn't follow `snake_case` for many of their definitions (e.g., ggproto uses names like GeomBar and argument formals use `na.rm` and `inherit.aes` in many, many places.)

PR #671, opened 2026-08-30
fix: Stop `lengths` rule from panicking on `sapply(FUN = length)`

Yousa-Mirage:
## Reproduce

A `test.R` like this:

``` r
x <- list(a = 1:2, b = 4:6)
x |> sapply(FUN = length)
#> a b 
#> 2 3

# The same as `x |> purrr::map_dbl(.f = length)`
```

will cause panic:

```bash
$jarl check test.R

thread 'main' (114726) panicked at crates/jarl-core/src/lints/base/lengths/lengths.rs:86:46:
```

And the auto-fix on this `test.R`:

``` r
x <- list(a = 1:2, b = 4:6)

sapply(X = x, FUN = length)
#> a b 
#> 2 3
sapply(, length)
#> Error in `sapply()`:
#> ! argument "X" is missing, with no default
```

result this below:

``` r
x <- list(a = 1:2, b = 4:6)

lengths(X = x)  # ! unused argument (X = x)
lengths()  # Shouldn't be fixed
```

## Change

Use `let arg_x = unwrap_or_return_none!(arg_x.and_then(|arg| arg.value()));` instead of `arg_x.unwrap()`.

Now `x |> sapply(FUN = length)` will not panic. And the auto-fix will result this:

``` r
x <- list(a = 1:2, b = 4:6)

lengths(x)  # ✅
sapply(, length)  # No change because of another error
```

### DIFF
--- a/crates/jarl-core/src/lints/base/lengths/lengths.rs
+++ b/crates/jarl-core/src/lints/base/lengths/lengths.rs
@@ -77,13 +77,14 @@ pub fn lengths(ast: &RCall, fn_name: &str) -> anyhow::Result<Option<Diagnostic>>
             .text_trimmed()
             == "length"
     {
+        let arg_x = unwrap_or_return_none!(arg_x.and_then(|arg| arg.value()));
         let range = ast.syntax().text_trimmed_range();
         let diagnostic = Diagnostic::new(
             Lengths,
             range,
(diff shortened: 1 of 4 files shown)


etiennebacher (approved):
Thanks!

PR #672, opened 2026-08-30
fix: Ensure `--fix` works with multibyte characters

Yousa-Mirage:
Before:

```r
# jarl-ignore seq: 原因一
x <- 1
# jarl-ignore browser: 原因二
y <- 2
```

After `--fix`:

```r
x <- 1
# jarl
```

Before:

```r
排序 <- 分数[order(分数)]
any(is.na("数据"))
```

`--fix` panicked:

```bash
$jarl check test.R --fix

thread 'main' (229944) panicked at /rustc/88d9e12ae178fab0fb5cc050a94da85685d449ea/library/core/src/slice/index.rs:1009:51:
range end index 47 out of range for slice of length 44
```

Change the content length difference before and after repair to be calculated based on UTF-8 bytes, and update the subsequent repair interval with byte displacement to avoid multi-byte characters (like Chinese) causing misalignment, data corruption or panic.

By the way, I'm not sure where to put these tests🥲.

### DIFF
--- a/crates/jarl-core/src/fix.rs
+++ b/crates/jarl-core/src/fix.rs
@@ -20,9 +20,6 @@ pub fn apply_fixes(fixes: &[Diagnostic], contents: &str) -> String {
     // length.
     let mut last_original_end: usize = 0;
 
-    // fix.start()/end() are byte offsets into the source, so the shift applied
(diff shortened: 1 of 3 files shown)


etiennebacher:
Thanks!

PR #676, opened 2026-09-01
fix: Correctly position suppression quickfixes with non-ASCII text

Yousa-Mirage:
Fixed the issue where the suppression quickfix insertion position was misaligned and might damage the code when LSP encountered non-ASCII characters at the beginning in UTF-16 encoding.

图片

will result:

```r
x <- "中文"; any(
               # jarl-ignore any_is_na: <reason>
               is.na(y))
```

if this file is UTF-16 rather than UTF-8.

This is indeed a very marginal coding problem that I don't think anyone will meet it. Since there was very little non-ASCII test content in the project, I asked AI to help analyze some issues that might trigger the boundaries of non-ASCII characters like Chinese.


etiennebacher:
Confirmed locally, thanks!

PR #677, opened 2026-09-01
fix: Prevent nested-call fixes from dropping extra unnamed arguments

Yousa-Mirage:
Before `test.R`:

```r
any(is.na(x), is.na(y))
any(duplicated(x), duplicated(y))
stop(paste0("a"), "b")
```

After fix with `jarl check test.R --fix --select=any_is_na,any_duplicated,condition_message`:

```r
anyNA(x)
anyDuplicated(x) > 0
stop("a")
```

This obviously breaks the code.

Nested-call autofixes for `any_is_na`, `any_duplicated`, and `condition_message` could silently drop extra unnamed arguments while replacing the entire outer call. The root cause was `get_nested_functions_content` selecting only the first unnamed argument without verifying that it was the only one.

### DIFF
--- a/crates/jarl-core/src/lints/base/condition_message/mod.rs
+++ b/crates/jarl-core/src/lints/base/condition_message/mod.rs
@@ -13,8 +13,6 @@ mod tests {
     fn test_no_lint_condition_message() {
         expect_no_lint("stop('boom')", "condition_message", None);
         expect_no_lint("stop('hello', 'there')", "condition_message", None);
-        expect_no_lint("stop(paste0('a'), 'b')", "condition_message", None);
-        expect_no_lint("warning(paste0('a'), 'b')", "condition_message", None);
         expect_no_lint("stop('boom', call. = FALSE)", "condition_message", None);
         expect_no_lint(
             "stop('hello', call. = FALSE, 'there')",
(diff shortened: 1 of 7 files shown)


crates/jarl-core/src/lints/base/condition_message/mod.rs
```diff
@@ -78,19 +80,6 @@ mod tests {
...
-        warning: condition_message
-         --> <test>:1:1
-          |
-        1 | stop(paste0('hello ', 'there'), call. = FALSE, ' again')
-          | -------------------------------------------------------- `stop(paste0(...))` can be simplified.
-          |
-          = help: Use `stop(...)` instead.
-        Found 1 error.
-        "
-        );
```
etiennebacher, line 93: I think this lint is correct, it's just the fix that is wrong. The fix should not drop the extra arguments:

``` r
stop(paste0('hello ', 'there'), call. = FALSE, ' again')
#> Error:
#> ! hello there again
stop('hello ', 'there', call. = FALSE, ' again')
#> Error:
#> ! hello there again
```

etiennebacher (changes requested):
Thanks! I agree for the changes to `any_is_na` and `any_duplicated`, but I think we can still report some lints and use some autofixes in `condition_call`.

etiennebacher:
We could also fix `any(is.na(x), is.na(y))` into `anyNA(c(x, y))` but this can considered a future improvement and left aside for now.

Yousa-Mirage:
I think there are some other issues with `condition_message`, but they have nothing to do with the problem this PR wants to fix.

> We could also fix `any(is.na(x), is.na(y))` into `anyNA(c(x, y))` but this can considered a future improvement and left aside for now.

Yes, maybe later.

etiennebacher (approved):
Thank you!

PR #678, opened 2026-09-01
fix: Avoid invalid `literal_coercion` fixes for strings with quotes

Yousa-Mirage:
Before `test.R`:

```r
as.character('say "hi"')
chr('say "hi"')
```

After fix with `--select=literal_coercion`:

```r
"say "hi""
"say "hi""
```

The code is broken.

I recovered the correct quote in the fix by recording the quote used in `String { content: String, quote: char }`.


etiennebacher (approved):
Great, thanks!

PR #679, opened 2026-09-01
chore: Update `annotate_snippets` to 0.12

etiennebacher:


crates/jarl/tests/integration/edge_cases.rs
```diff
@@ -92,7 +92,7 @@ fn test_jarl_with_tabs_on_earlier_lines() -> anyhow::Result<()> {
     exit_code: 1
     ----- stdout -----
     warning: any_is_na
-     --> test.R:4:5
+     --> test.R:4:2
```
etiennebacher, line 95: Checked locally that this change is correct, previously it would put the cursor in the middle of `any(is.na(x))`.

PR #680, opened 2026-09-02
fix: Stop `expect_type` rule from duplicating `expect_null` lints

Yousa-Mirage:
`test.R`:

```r
expect_true(is.null(x))
```

Then run check:
```bash
$jarl check test.R --select=TESTTHAT

warning: expect_null
 --> test.R:1:1
  |
1 | expect_true(is.null(x))
  | ----------------------- `expect_true(is.null(x))` is not as clear as `expect_null(x)`.
  |
  = help: Use `expect_null(x)` instead.

warning: expect_type
 --> test.R:1:1
  |
1 | expect_true(is.null(x))
  | ----------------------- `expect_true(is.<t>(x))` can be hard to read.
  |
  = help: Use `expect_type(x, t)` instead.


Found 2 errors.
2 fixable with the `--fix` option.
```

Actually it's the same problem. But it was reported twice by `expect_null` and `expect_type`.

I removed the check in `expect_type`. And `expect_equal(typeof(x), "NULL")` will still be reported by `expect_type` rule.


etiennebacher (approved):
Thanks!

PR #681, opened 2026-09-02
fix: Match `dplyr_filter_out` guards by identifier

Yousa-Mirage:
Before: 
```r
library(dplyr)

data <- tibble(
  row = 1:3,
  x = c(NA_real_, 1, 1),
  x2 = c(-1, -1, 1)
)

data |> filter(x2 > 0 | is.na(x))
#> # A tibble: 2 × 3
#>     row     x    x2
#>   <int> <dbl> <dbl>
#> 1     1    NA    -1
#> 2     3     1     1
```

After `jarl check test.R --fix --select=DPLYR`:

```r
data |> filter_out(x2 <= 0)
#> # A tibble: 1 × 3
#>     row     x    x2
#>   <int> <dbl> <dbl>
#> 1     3     1     1
```

The `is.na()` guard must match a complete identifier, not a substring `!cond_text.contains(&is_na_arg)`.


etiennebacher (approved):
Oops, thanks!

PR #683, opened 2026-09-02
fix: Avoid invalid `pipe_consistency` fixes for bare RHS

Yousa-Mirage:
Before:

```r
x %>% sum
```

After `jarl check test.R --select pipe_consistency --fix --unsafe-fixes --min-r-version 4.2`:

```r
x |> sum
```

The code is broken.

### DIFF
--- a/crates/jarl-core/src/lints/base/pipe_consistency/pipe_consistency.rs
+++ b/crates/jarl-core/src/lints/base/pipe_consistency/pipe_consistency.rs
@@ -109,10 +109,23 @@ pub fn pipe_consistency(
     };
 
     let bin_range = ast.syntax().text_trimmed_range();
+    let op_range = operator.text_trimmed_range();
+
+    if preferred_is_base && right.as_r_call().is_none() {
+        return Ok(Some(Diagnostic::new(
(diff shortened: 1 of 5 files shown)


crates/jarl-core/src/lints/base/pipe_consistency/pipe_consistency.rs
```diff
@@ -109,10 +109,23 @@ pub fn pipe_consistency(
...
+        return Ok(Some(Diagnostic::new(
+            ViolationData::new(
+                Rule::PipeConsistency,
+                body.to_string(),
+                Some(suggestion.to_string()),
+            ),
+            op_range,
+            Fix::empty(),
+        )));
+    }
```
etiennebacher: I think there could be a fix to add `()` where needed then?

etiennebacher (commented):
Thanks, I thought I had caught that but maybe I was just lazy and it was one of the reasons I marked this fix as unsafe.

Yousa-Mirage:
OK, now `x %>% sum` will be fixed to `x |> sum()`.

PR #687, opened 2026-09-03
fix: Avoid panic on missing sample arguments

Yousa-Mirage:
Before `test.R`:

```r
sample(, 1)
```

will cause panic:

```bash
$jarl check test.R --select sample_int

thread 'main' (20586) panicked at crates/jarl-core/src/lints/base/sample_int/sample_int.rs:61:33:
```

A minor modification.


etiennebacher (approved):
Nice catch!

PR #688, opened 2026-09-03
fix: Prevent `outer_negation` from rewriting `+` and `-` operators

Yousa-Mirage:
Before `test.R`:

```r
x <- c(TRUE, TRUE)

any(+x)
#> TRUE
any(-x)
#> TRUE
```

After fix:

```r
x <- c(TRUE, TRUE)

!all(x)
#> FALSE
!all(x)
#> FALSE
```

This is because the current code treats `+` and `-` as unary operators like `!` and applies the same fix to them.


etiennebacher (approved):
Perfect, thanks!

PR #691, opened 2026-09-03
fix: Tweak the message of `unused_function`

### LINKED ISSUE #682: unused_function when developing a package
Not entirely sure this is a false positive as such, but I think the warning text could do with amending.

I was editing a personal package of simple helper functions to add a new one and noticed I got the following warning from Jarl:

`` `bytes` is defined but never called in this package. Defined at R/bytes.R:1:1 but never called``

For reference the function is a simple wrapper around `pretty_bytes()` and `object.size()`:

```r
bytes <- function(x) {
  prettyunits::pretty_bytes(object.size(x))
}
```

This function is isn't going to be called within the package since it's not an internal function and is intended to be exported. After adding a roxygen skeleton with an `@export` tag the Jarl warning persists, I thought that might cause it to disappear. The warning disappears only after I've run `devtools::document()`, I presume because Jarl looks at the NAMESPACE and sees it's listed as an exported package.

The warning text therefore feels a bit off, as it reads to me like Jarl considers the function internal and I don't think of the NAMESPACE as "calling" a function.

I'm not sure how easy it is to scan for roxygen tags, in which case perhaps the warning text should be amended to make it clearer that the warning will disappear once the function appears in the NAMESPACE. Something like:

``
`bytes` is defined but is not called or documented in this package. Defined at R/bytes.R:1:1 but never called and not listed in NAMESPACE.
``

mattkerlogue:
Fixes #682, providing clearer warning/message when a function intended to be exported has not yet been listed in `NAMESPACE`.

Warning message for unused functions now reads:

`` `{name}` is defined but is never called in this package nor is it exported in NAMESPACE. Defined at {path}:{line}:{col}.``

Documentation also updated as follows:

```
While exported functions may not be called by other functions, they must be
declared in the package's NAMESPACE file before they are visible to users.
```

`cargo insta` snapshots have been re-run, all tests in `cargo test` passing.

`just lint` and `just document` were also run.

### DIFF
--- a/crates/jarl-core/src/lints/base/unused_function/unused_function.rs
+++ b/crates/jarl-core/src/lints/base/unused_function/unused_function.rs
@@ -24,6 +24,9 @@ use crate::package::{FileScope, SharedFileData};
 /// refactoring. Removing it keeps the codebase easier to understand and
 /// maintain.
 ///
+/// While exported functions may not be called by other functions, they must be
+/// declared in the package's NAMESPACE file before they are visible to users.
+///
 /// ## Limitations
 ///
 /// There are many ways to call a function in R code (e.g. `foo()`,
@@ -57,6 +60,7 @@ use crate::package::{FileScope, SharedFileData};
 /// # `check_length()` isn't exported but and isn't used anywhere, so it is
 /// # reported.
 /// ```
+///
 // ## Implementation
 //
 // Operates on the already-scanned `SharedFileData` of a package rather than
@@ -197,7 +201,7 @@ pub(crate) fn compute_unused_from_shared(
 
                 if occurrences <= definitions && !extra_symbol_set.contains(name.as_str()) {
                     let help = format!(
-                        "Defined at {path}:{line}:{col} but never called",
+                        "Defined at {path}:{line}:{col}.",
                         path = file.rel_path.display()
                     );
                     unused.push((name.clone(), *range, help));
--- a/crates/jarl-core/src/lints/base/unused_function/unused_function.rs
+++ b/crates/jarl-core/src/lints/base/unused_function/unused_function.rs
@@ -20,12 +20,10 @@ use crate::package::{FileScope, SharedFileData};
 ///
 /// ## Why is this bad?
 ///
-/// An internal function that is never called is likely dead code left over from
-/// refactoring. Removing it keeps the codebase easier to understand and
-/// maintain.
-///
-/// While exported functions may not be called by other functions, they must be
-/// declared in the package's NAMESPACE file before they are visible to users.
+/// Functions must be documented in NAMESPACE to be exported to end users. A
+/// unction that is never called nor exported is likely dead code left over
+/// from refactoring. Removing unused internal functions keeps the codebase
+/// easier to understand and maintain.
 ///
 /// ## Limitations
 ///
--- a/crates/jarl-core/src/lints/base/unused_function/unused_function.rs
+++ b/crates/jarl-core/src/lints/base/unused_function/unused_function.rs
@@ -21,7 +21,7 @@ use crate::package::{FileScope, SharedFileData};
 /// ## Why is this bad?
 ///
 /// Functions must be documented in NAMESPACE to be exported to end users. A
-/// unction that is never called nor exported is likely dead code left over
+/// function that is never called nor exported is likely dead code left over
 /// from refactoring. Removing unused internal functions keeps the codebase
 /// easier to understand and maintain.
 ///
@@ -58,7 +58,6 @@ use crate::package::{FileScope, SharedFileData};
 /// # `check_length()` isn't exported but and isn't used anywhere, so it is
 /// # reported.
 /// ```
-///
 // ## Implementation
 //
 // Operates on the already-scanned `SharedFileData` of a package rather than
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -2,7 +2,6 @@ project:
   type: website
   pre-render: copy_readme.R
 filters:
-- linkify-github-refs.lua
 - newpagelink.lua
 format:
   html:
--- a/crates/jarl-core/src/analyze/document.rs
+++ b/crates/jarl-core/src/analyze/document.rs
@@ -133,7 +133,7 @@ pub(crate) fn check_document(
             checker.report_diagnostic(Some(Diagnostic::new(
                 ViolationData::new(
                     Rule::UnusedFunction,
-                    format!("`{name}` is defined but never called in this package."),
+                    format!("`{name}` is defined but is never called in this package nor is it exported in NAMESPACE."),
                     Some(help.clone()),
                 ),
                 *range,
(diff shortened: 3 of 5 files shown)


crates/jarl-core/src/lints/base/unused_function/unused_function.rs
```diff
@@ -24,6 +24,9 @@ use crate::package::{FileScope, SharedFileData};
 /// refactoring. Removing it keeps the codebase easier to understand and
 /// maintain.
 ///
+/// While exported functions may not be called by other functions, they must be
+/// declared in the package's NAMESPACE file before they are visible to users.
+///
```
etiennebacher: I don't think this phrasing is correct because being declared in `NAMESPACE` is the condition to be exported (put differently, you cannot have an exported function not declared in `NAMESPACE`).

I would rather remove this paragraph and change the paragraph above instead:

```diff
- /// An internal function that is never called is likely dead code left over from
- /// refactoring. Removing it keeps the codebase easier to understand and
+ /// An internal function that is never called and not exported is likely dead code 
+ /// left over from refactoring. Removing it keeps the codebase easier to understand and
```

docs/_quarto.yml
```diff
@@ -2,7 +2,6 @@ project:
   type: website
   pre-render: copy_readme.R
 filters:
-- linkify-github-refs.lua
```
etiennebacher: Why was this line removed?

etiennebacher (changes requested):
Thank you! Just a couple of comments to address.

You can ignore the failing CI workflow about the changelog.

crates/jarl-core/src/lints/base/unused_function/unused_function.rs
mattkerlogue: My reasoning was more aimed at helping novice package developers understand the rationale getting a warning. In particular when the unused_function rule is causing a warning while they're still in development and haven't yet added the function to `NAMESPACE`, especially if they're using `{devtools}`/`{roxygen2}` to handle documentation and managing the `NAMESPACE` and haven't called `devtools::document()` yet.

This is why I referenced the `@export` roxygen tag in #682. It can be very common that an exported function is not called by any other functions within the package. But if you've already added the `@export` tag then you've done the principal mental load to declare the function as due for export, even if you've not yet done the practical step to actually get it written into the `NAMESPACE`.

There's many a time I've added the `@export` tag and then forgotten to call `document()`. Or to put it another way: the function has gone through passport control and is waiting in the airport terminal, it just hasn't yet boarded the flight yet.

How about:

Functions must be documented in `NAMESPACE` to be exported to end users. An internal function that is never called is likely dead code left over from refactoring. Removing unused internal functions keeps the codebase easier to understand and maintain.

docs/_quarto.yml
mattkerlogue: Honestly, I don't recall removing it and I didn't notice it when staging the commit.

On further investigation it seems to be highly bizarre. I switched to the main branch, ran `cat docs/_quarto.yml` in terminal and this shows the `linkify-github-refs.lua` filter there in the text.

However I then opened VS Code to check the file and it's somehow marked up as deleting that line.

Screenshot 2026-09-08 at 15 32 07

Re-doing `cat docs/_quarto.yml` in terminal confirms the filter has been removed, without me having done anything. 

Will re-instate.

One thing I did notice in checking this though is that the indentation in the YAML file seems to be a bit off, do you want me to fix? Can do so in this branch or a separate branch.

```yml
project:
  type: website
  pre-render: copy_readme.R
filters:
- linkify-github-refs.lua
- newpagelink.lua
format:
  html:
    theme: morph
    css:
    - styles.css
    - dark-mode.css
```

Will work, but YAML spec expects:

```yml
project:
  type: website
  pre-render: copy_readme.R
filters:
  - linkify-github-refs.lua
  - newpagelink.lua
format:
  html:
    theme: morph
    css:
      - styles.css
      - dark-mode.css
```

etiennebacher: Thanks, I made the YAML changes in https://github.com/etiennebacher/jarl/commit/d1759a4086f462c567fbd59341051b843aeb97d1 on `main`, so there's a git conflict to resolve. The deleted file still needs to be reverted.

crates/jarl-core/src/lints/base/unused_function/unused_function.rs
etiennebacher: > How about:
> 
> Functions must be documented in `NAMESPACE` to be exported to end users. An internal function that is never called is likely dead code left over from refactoring. Removing unused internal functions keeps the codebase easier to understand and maintain.

Yup, sounds good to me, just with the small following change:

> Functions must be documented in `NAMESPACE` to be exported to end users. **A function that is never called nor exported** is likely dead code left over from refactoring. Removing unused internal functions keeps the codebase easier to understand and maintain.

(emphasis added just to highlight the change, it's not needed in the docs)

Are you fine with that?

mattkerlogue: Sounds good to me. Will implement tomorrow.

docs/_quarto.yml
mattkerlogue: Ah, cool. Will resolve tomorrow.

mattkerlogue:
Have made the changes to the documentation, resolved the quarto.yml issue and merged in main.

`cargo test` reported one failure but I couldn't diagnose the point of failure, but don't think its related to any of the changes I made.

```
thread 'package_cache::tests::test_run_rscript_for_pkg_info_integration' (989011) panicked at crates/jarl-core/src/package_cache.rs:510:42:
```

Running with a backtrace did not give any further details.

mattkerlogue:
I think I've uncovered more of what was happening with `_quarto.yml`. I'll start a new issue to explain in a separate thread.

crates/jarl-core/src/lints/base/unused_function/unused_function.rs
```diff
@@ -20,9 +20,10 @@ use crate::package::{FileScope, SharedFileData};
 ///
 /// ## Why is this bad?
 ///
-/// An internal function that is never called is likely dead code left over from
-/// refactoring. Removing it keeps the codebase easier to understand and
-/// maintain.
+/// Functions must be documented in NAMESPACE to be exported to end users. A
+/// unction that is never called nor exported is likely dead code left over
```
etiennebacher:
```suggestion
/// function that is never called nor exported is likely dead code left over
```

crates/jarl-core/src/lints/base/unused_function/unused_function.rs
```diff
@@ -57,6 +58,7 @@ use crate::package::{FileScope, SharedFileData};
 /// # `check_length()` isn't exported but and isn't used anywhere, so it is
 /// # reported.
 /// ```
+///
```
etiennebacher:
```suggestion
```

etiennebacher (changes requested):
Thanks! Please include the two fixes below and regenerate the md file

etiennebacher (approved):
Ah I had tried to push directly to this branch but couldn't figure out where the commit went 😄

Issue #689, opened 2026-09-03
Possible refactors depending on the R version


etiennebacher:
From https://github.com/r-lib/usethis/issues/2220#issuecomment-5458128346

---

Tidyverse helpers obviated by recent base R releases:

**R 4.0.0**
- [x] Drop `stringsAsFactors = FALSE` — it is now the default in `data.frame()` and `read.table()`
- [ ] Use raw character constants `r"(...)"` for strings with many `\` escapes (e.g. regexes) instead of doubled backslashes
- [ ] Use `deparse1()` instead of `paste(deparse(x), collapse = " ")`
- [ ] Use `list2DF()` to create a data frame from a list instead of `as.data.frame()`/`tibble()` workarounds

**R 4.1.0**
- [x] Switch to the base pipe `|>` instead of magrittr's `%>%`
- [ ] Replace purrr `~` lambdas and inline `function()` with the `\(x)` shorthand
- [ ] Use `...names()` instead of `names(list(...))`

**R 4.2.0**
- [ ] Use the `_` placeholder with `|>` to insert the lhs into a named argument (e.g. `x |> lm(y ~ x, data = _)`) instead of magrittr's `.` or an anonymous function

**R 4.3.0**
- [ ] Use the `_` placeholder in extraction calls on the rhs of `|>` (e.g. `x |> _$coef` or `x |> _$coef[[2]]`) instead of purrr-style lambdas like `~ .$coef`
- [ ] Drop `useBytes = TRUE` in `grep()`/`sub()`/`gsub()`/`strsplit()` — performance was improved for `perl = TRUE`/`fixed = TRUE` and `useBytes` is now discouraged

**R 4.4.0**
- [ ] Use base `%||%` instead of rlang's `%||%`
- [ ] Use `sort_by()` for simple data frame row sorting instead of `dplyr::arrange()`

**R 4.5.0**
- [x] Use `grepv()` instead of `grep(..., value = TRUE)` (and as a base analogue of `stringr::str_subset()`)

**R 4.6.0**
- [x] Use base `%notin%` instead of a home-grown `%nin%` or `Negate(\`%in%\`)`
- [ ] Use `match.arg(several.ok = "all")` instead of rlang-style multi-value argument matching helpers
- [ ] Use `substring(x, i1, i2 = NULL)` ("to end of string") instead of `stringr::str_sub()` workarounds

maelle:
Related: if I want to see all the rules related to `R 4.1.0`, how do I do that?

[...]

PR #694, opened 2026-09-04
feat: Add `deparse1`

maelle:
Co-Authored-By: Claude Sonnet 5

Part of #689

### DIFF
--- a/crates/jarl-core/src/lints/base/deparse1/deparse1.rs
+++ b/crates/jarl-core/src/lints/base/deparse1/deparse1.rs
@@ -26,8 +26,9 @@ impl Violation for Deparse1 {
 }
 
 pub fn deparse1(ast: &RCall, fn_name: &str) -> anyhow::Result<Option<Diagnostic>> {
-    let (inner_content, outer_syntax) =
-        unwrap_or_return_none!(get_nested_functions_content(ast, fn_name, "paste", "deparse")?);
+    let (inner_content, outer_syntax) = unwrap_or_return_none!(get_nested_functions_content(
+        ast, fn_name, "paste", "deparse"
+    )?);
 
     let collapse = unwrap_or_return_none!(get_arg_by_name(&ast.arguments()?.items(), "collapse"));
     let collapse_value = unwrap_or_return_none!(collapse.value());
--- a/crates/jarl-core/src/lints/base/deparse1/deparse1.rs
+++ b/crates/jarl-core/src/lints/base/deparse1/deparse1.rs
@@ -1,0 +1,54 @@
+use crate::diagnostic::*;
+use crate::rule_set::Rule;
+use crate::utils::{get_arg_by_name, get_nested_functions_content, node_contains_comments};
+use air_r_syntax::*;
+use oak_core::syntax_ext::RStringValueExt;
+
+pub struct Deparse1;
+
+/// Version added: 0.6.1
+///
+/// ## What it does
+///
+/// Checks for usage of `paste(deparse(x), collapse = " ")`.
+///
+/// [...]
+impl Violation for Deparse1 {
+    fn rule(&self) -> Rule {
+        Rule::Deparse1
+    }
+    fn body(&self) -> String {
+        "`paste(deparse(x), collapse = \" \")` is inefficient and can be hard to read.".to_string()
+    }
+    fn suggestion(&self) -> Option<String> {
+        Some("Use `deparse1(x)` instead.".to_string())
+    }
+}
+
+pub fn deparse1(ast: &RCall, fn_name: &str) -> anyhow::Result<Option<Diagnostic>> {
+    let (inner_content, outer_syntax) =
+        unwrap_or_return_none!(get_nested_functions_content(ast, fn_name, "paste", "deparse")?);
+
+    let collapse = unwrap_or_return_none!(get_arg_by_name(&ast.arguments()?.items(), "collapse"));
+    let collapse_value = unwrap_or_return_none!(collapse.value());
+
+    let collapse_r_value = unwrap_or_return_none!(collapse_value.as_any_r_value());
+    let collapse_string = unwrap_or_return_none!(collapse_r_value.as_r_string_value());
+
+    if collapse_string.string_text().as_deref() != Some(" ") {
+        return Ok(None);
+    }
+
+    let range = outer_syntax.text_trimmed_range();
+    let diagnostic = Diagnostic::new(
+        Deparse1,
+        range,
+        Fix::new(
+            range,
+            format!("deparse1({inner_content})"),
+            node_contains_comments(&outer_syntax),
+        ),
+    );
+
+    Ok(Some(diagnostic))
+}
--- a/crates/jarl-core/src/lints/base/deparse1/deparse1.rs
+++ b/crates/jarl-core/src/lints/base/deparse1/deparse1.rs
@@ -12,7 +12,31 @@ pub struct Deparse1;
 ///
 /// Checks for usage of `paste(deparse(x), collapse = " ")`.
 ///
-/// [...]
+/// ## Why is this bad?
+///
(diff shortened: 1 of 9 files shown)


crates/jarl-core/src/lints/base/deparse1/deparse1.rs
```diff
@@ -0,0 +1,54 @@
...
+
+pub struct Deparse1;
+
+/// Version added: 0.6.1
+///
+/// ## What it does
+///
+/// Checks for usage of `paste(deparse(x), collapse = " ")`.
+///
+/// [...]
```
maelle: I'm not sure what else I should add here, compared to the docs in `docs/rules/deparse1.md`

crates/jarl-core/src/lints/base/deparse1/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+#[cfg(test)]
+mod tests {
+    use crate::utils_test::*;
+    use insta::assert_snapshot;
+
+    fn snapshot_lint(code: &str) -> String {
+        format_diagnostics(code, "deparse1", Some("4.1.0"))
+    }
+
+    #[test]
```
maelle, line 12: https://github.com/r-lib/lintr/blob/2df1a7b337f2455bb18d6ee60899c08bce046686/tests/testthat/test-paste_linter.R#L315

crates/jarl-core/src/rule_set.rs
```diff
@@ -315,6 +315,13 @@ declare_rules! {
         fix: Safe,
         min_r_version: None,
     },
+    Deparse1 => {
+        name: "deparse1",
+        categories: [Perf, Read],
+        default: Enabled,
+        fix: Safe,
+        min_r_version: Some((4, 1, 0)),
```
etiennebacher: It was added in 4.0.0 (tests and examples also need to be updated)

crates/jarl-core/src/lints/base/deparse1/deparse1.rs
```diff
@@ -0,0 +1,55 @@
...
+impl Violation for Deparse1 {
+    fn rule(&self) -> Rule {
+        Rule::Deparse1
+    }
+    fn body(&self) -> String {
+        "`paste(deparse(x), collapse = \" \")` is inefficient and can be hard to read.".to_string()
```
etiennebacher, line 45:
```suggestion
        "`paste(deparse(x), collapse = \" \")` can be simplified.".to_string()
```

I know this sentence is used elsewhere and I should review these cases too, but I think this message is better.

crates/jarl-core/src/lints/base/deparse1/deparse1.rs
etiennebacher: I think something like:

```
# Why is it bad?

This pattern of code can be simplified by using `deparse1()`, 
added in R 4.0.0.

# Examples

<add an example>

# References

See `?deparse1`
```

is enough.

---

Edit: Ah I wrote that before seeing the `md` file 😄

crates/jarl-core/src/rule_set.rs
```diff
@@ -315,6 +315,13 @@ declare_rules! {
         fix: Safe,
         min_r_version: None,
     },
+    Deparse1 => {
+        name: "deparse1",
+        categories: [Perf, Read],
+        default: Enabled,
+        fix: Safe,
```
etiennebacher: Hmm, should it be an unsafe fix?

``` r
identical(
  deparse1(args(library)), 
  paste(deparse(args(library)), collapse = " ")
)
#> [1] FALSE
```

etiennebacher (changes requested):
Thanks, I haven't reviewed the implementation or tests yet, but already a few comments.

maelle:
Regarding the md, I realize I was confused by "add or update the rule page in docs/rules/<rule_name>.md". Does that md come from what is written in the rs? If so, the sentence should be clarified.

etiennebacher:
> Does that md come from what is written in the rs? 

Yes it is copied from there, I've added https://github.com/etiennebacher/jarl/commit/547b20d7569f72ac527bdaeeae3c678cee8244f1

crates/jarl-core/src/rule_set.rs
```diff
@@ -315,6 +315,13 @@ declare_rules! {
         fix: Safe,
         min_r_version: None,
     },
+    Deparse1 => {
+        name: "deparse1",
+        categories: [Perf, Read],
```
etiennebacher, line 320: I think this is just a "Read" rule, there's no perf gain.

crates/jarl-core/src/lints/base/deparse1/deparse1.rs
```diff
@@ -0,0 +1,79 @@
+use crate::diagnostic::*;
+use crate::rule_set::Rule;
+use crate::utils::{get_arg_by_name, get_nested_functions_content, node_contains_comments};
+use air_r_syntax::*;
+use oak_core::syntax_ext::RStringValueExt;
+
+pub struct Deparse1;
+
+/// Version added: 0.6.1
```
etiennebacher, line 9: You can't tell what the next version is going to be (cf one of your comments on a github issue) but it's probably gonna be 0.7.0.

crates/jarl-core/src/lints/base/deparse1/deparse1.rs
```diff
@@ -0,0 +1,79 @@
...
+        Deparse1,
+        range,
+        Fix::new(
+            range,
+            format!("deparse1({inner_content})"),
+            node_contains_comments(&outer_syntax),
+        ),
+    );
+
+    Ok(Some(diagnostic))
```
etiennebacher, line 78: Just out of curiosity I tried to convert this to more idiomatic Rust:

```rust
    if let Some(collapse) = get_arg_by_name(&ast.arguments()?.items(), "collapse")
        && let Some(value) = collapse.value()
        && let Some(r_value) = value.as_any_r_value()
        && let Some(collapse_string) = r_value.as_r_string_value()
        && collapse_string.string_text().as_deref() == Some(" ")
    {
        let range = outer_syntax.text_trimmed_range();
        let diagnostic = Diagnostic::new(
            Deparse1,
            range,
            Fix::new(
                range,
                format!("deparse1({inner_content})"),
                node_contains_comments(&outer_syntax),
            ),
        );

        return Ok(Some(diagnostic));
    }

    return Ok(None);
```
but in the end it takes as many lines so let's keep your approach.

crates/jarl-core/src/lints/base/deparse1/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+         --> <test>:3:3
+          |
+        3 |   paste(deparse(y), collapse = ' ')
+          |   --------------------------------- `paste(deparse(x), collapse = " ")` is inefficient and can be hard to read.
+          |
+          = help: Use `deparse1(x)` instead.
+        Found 2 errors.
+        "#
+        );
+    }
```
etiennebacher, line 158: No need for that. Instead, can you add some tests to handle comments? See in other rules how this is tested but basically it should ensure that fixes are applied if the comment is leading or trailing but not if it's somewhere inside the expression.

etiennebacher (changes requested):
Looks good overall, thanks! A few comments to address

crates/jarl-core/src/lints/base/deparse1/deparse1.rs
maelle: But isn't it better to have something idiomatic?

etiennebacher: I don't think non-idiomatic code matters in the rule definitions (`unwrap_or_return_none!()` itself is not really idiomatic but it's used everywhere).

Up to you in the end if you want to incorporate this change as you do other changes.

PR #696, opened 2026-09-05
fix: Avoid false positives for quote-only strings in `nzchar`

Yousa-Mirage:
See:

```r
# jarl check test.R --select nzchar

x == "'"  # will report `x == ""`
x != "'"  # will report `x != ""`
x == "''"  # will report `x == ""`
x != "''"  # will report `x != ""`
```

I use `oak_core::syntax_ext::RStringValueExt::string_text()` to get the string contents without surrounding quotes, instead of cleaning quotes manually.

Furthermore, I believe this rule should also check for patterns like `nchar(x) == 0`, consistent with [lintr's `nzchar_linter`](https://lintr.r-lib.org/reference/nzchar_linter.html). I noted your comment at https://github.com/etiennebacher/jarl/pull/406#discussion_r2964881347 stating that *"it's not part of the rule"*, but I feel this is essentially the same issue and that both cases should be replaced with `nzchar()`.

What's your opinion @etiennebacher ?

---

`lintr` checks all comparison expressions here:

```r
comparison_msg_map <- c(
  GT = 'Use nzchar(x) instead of x > "". ',
  NE = 'Use nzchar(x) instead of x != "". ',
  LE = 'Use !nzchar(x) instead of x <= "". ',
  EQ = 'Use !nzchar(x) instead of x == "". ',
  GE = 'x >= "" is always true, maybe you want nzchar(x)? ',
  LT = 'x < "" is always false, maybe you want !nzchar(x)? '
)

nchar_msg_map <- c(
  GT = "Use nzchar(x) instead of nchar(x) > 0. ",
  NE = "Use nzchar(x) instead of nchar(x) != 0. ",
  LE = "Use !nzchar(x) instead of nchar(x) <= 0. ",
  EQ = "Use !nzchar(x) instead of nchar(x) == 0. ",
  GE = "nchar(x) >= 0 is always true, maybe you want nzchar(x)? ",
  LT = "nchar(x) < 0 is always false, maybe you want !nzchar(x)? "
)
```

### DIFF
--- a/crates/jarl-core/src/lints/base/nzchar/mod.rs
+++ b/crates/jarl-core/src/lints/base/nzchar/mod.rs
@@ -40,11 +40,26 @@ mod tests {
         );
 
         assert_snapshot!(
+            snapshot_lint(r#"x == r"()""#),
+            @r#"
+        warning: nzchar
+         --> <test>:1:1
+          |
+        1 | x == r"()"
+          | ---------- `x == ""` is inefficient.
+          |
+          = help: Use `!nzchar(x)` instead.
+        Found 1 error.
+        "#
+        );
+
+        assert_snapshot!(
             "fix_output",
             get_unsafe_fixed_text(
                 vec![
                     "x == ''",
                     "x != ''",
+                    r#"x == r"()""#,
                     "foo(x(y)) == ''",
                     "'' == x",
                     "which(c(a, b, c) == '')"
--- a/crates/jarl-core/src/lints/base/nzchar/mod.rs
+++ b/crates/jarl-core/src/lints/base/nzchar/mod.rs
@@ -62,6 +62,14 @@ mod tests {
         expect_no_lint("x %in% ''", "nzchar", None);
 
         expect_no_lint("x + ''", "nzchar", None);
+
+        expect_no_lint(r#"x == "'"#, "nzchar", None);
+
+        expect_no_lint(r#"x != "'"#, "nzchar", None);
+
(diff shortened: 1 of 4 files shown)


etiennebacher:
Thanks!

> Furthermore, I believe this rule should also check for patterns like nchar(x) == 0, consistent with [lintr's nzchar_linter](https://lintr.r-lib.org/reference/nzchar_linter.html). I noted your comment at https://github.com/etiennebacher/jarl/pull/406#discussion_r2964881347 stating that "it's not part of the rule", but I feel this is essentially the same issue and that both cases should be replaced with nzchar().

I'm fine with adding checks for `nchar(x) == 0`. I think my reluctance was that `!nzchar(x)` is really less readable, but this rule is actually in the "perf" category only so it is fine. For future readers of this thread, `nzchar()` is more performant on this case too:

``` r
x <- sample(c(countrycode::codelist$country.name.en, ""), 1e6, TRUE)

bench::mark(
  nchar(x) == 0,
  !nzchar(x)
)
#> # A tibble: 2 × 6
#>   expression         min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>    <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 nchar(x) == 0  31.04ms  31.45ms      31.8    7.63MB     14.5
#> 2 !nzchar(x)      1.83ms   1.93ms     491.     7.63MB    237.
```

Yousa-Mirage:
OK. I'll create another PR to check for `nchar(x) == 0`.

PR #697, opened 2026-09-05
fix: Traverse namespace, extractions, and selector nodes

Yousa-Mirage:
See:

```r
# jarl check test.R --select quotes

df['col']
foo('arg')
df$'col'  # No lint
df@'slot'  # No lint
pkg::'fn'()  # No lint
('fn')()  # No lint
(function() 'body')()  # No lint
```

The expression walker skipped extract and call-position subtrees, so quoted selectors were never passed to the `quotes` rule.


etiennebacher (approved):
Thanks, good catch!

I directly edited this PR to be a bit more general than the original focus on the "quotes" rule.

Yousa-Mirage:
Thank you for these changes. I will take a look to learn.

PR #698, opened 2026-09-05
fix: Guard `string_boundary` fixes against mismatched widths

Yousa-Mirage:
Refactored the `string_boundary` rule to fix some bugs. Currently no human review and optimization so draft PR now.

The current code is generated by GPT-6 and seems to be of very good quality. It even fixed the problem of missing parameters that I didn't notice at the same time. The cost was that a single task almost reached my 5-hour limit.


Yousa-Mirage:
OK. I've reviewed this PR and I think it's ready for review. This PR has a lot of changes, but I think the modified `string_boundary.rs` is clear to read and understand. Thank you for review.

PR #705, opened 2026-09-10
feat: Detect zero comparisons of `nchar()` in the `nzchar` rule

Yousa-Mirage:
See the discussion at #696. Now `nzchar` rule will check for comparisons of `nchar(x)` with zero, such as `nchar(x) == 0`.

### DIFF
--- a/crates/jarl-core/src/lints/base/nzchar/nzchar.rs
+++ b/crates/jarl-core/src/lints/base/nzchar/nzchar.rs
@@ -1,16 +1,18 @@
 use crate::diagnostic::*;
 use crate::rule_set::Rule;
-use crate::utils::node_contains_comments;
+use crate::utils::{Formals, get_arg, get_function_name, node_contains_comments};
 use air_r_syntax::*;
-use biome_rowan::AstNode;
+use biome_rowan::{AstNode, AstSeparatedList};
 use jarl_semantic::strings::get_string_literal_contents;
+
+const FORMALS_NCHAR: Formals = &["x"];
 
 /// Version added: 0.5.0
 ///
 /// ## What it does
 ///
-/// Checks for usage of `x != ""` or `x == ""`
-///  instead of `nzchar(x)` or `!nzchar(x)`.
+/// Checks for usage of `x != ""` or `x == ""`, or comparisons of `nchar(x)`
+/// with zero, such as `nchar(x) == 0`, instead of `nzchar(x)` or `!nzchar(x)`.
 ///
 /// ## Why is this bad?
 /// `x == ""` is less efficient than `!nzchar(x)`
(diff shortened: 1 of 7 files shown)


crates/jarl-core/src/lints/base/nzchar/nzchar.rs
```diff
@@ -107,3 +115,133 @@ pub fn nzchar(ast: &RBinaryExpression) -> anyhow::Result<Option<Diagnostic>> {
...
+        if let Some(argument) =
+            get_arg(call, FORMALS_NCHAR, "x").and_then(|argument| argument.value())
+        {
+            let range = ast.syntax().text_trimmed_range();
+            let argument = argument.to_trimmed_string();
+            let (body, suggestion, replacement) = match operator {
+                RSyntaxKind::GREATER_THAN => (
+                    "`nchar(x) > 0` is inefficient.",
+                    "Use `nzchar(x)` instead.",
+                    Some(format!("nzchar({argument})")),
```
etiennebacher, line 172: All fixes for this rule (including those before this PR) are unsafe, but actually I think we just need to add `keepNA = TRUE` to make them safe:

``` r
x <- sample(c(letters, "", NA), 100, TRUE)

waldo::compare(nchar(x) == 0, !nzchar(x))
#> `old[60:66]`: FALSE FALSE FALSE <NA>  FALSE FALSE FALSE
#> `new[60:66]`: FALSE FALSE FALSE FALSE FALSE FALSE FALSE
#> 
#> `old[69:75]`: FALSE FALSE FALSE <NA>  FALSE FALSE FALSE
#> `new[69:75]`: FALSE FALSE FALSE FALSE FALSE FALSE FALSE
#> 
#> `old[76:82]`: FALSE FALSE FALSE <NA>  FALSE FALSE FALSE
#> `new[76:82]`: FALSE FALSE FALSE FALSE FALSE FALSE FALSE
waldo::compare(nchar(x) == 0, !nzchar(x, keepNA = TRUE))
#> ✔ No differences
```

This is specified in the existing docs but I don't know why it wasn't added from the start. It's bad for readability but this is a "perf" rule anyway. Can you make the change either in this PR or in a later one?

We don't report `nchar()` when args other than `x` are specified so we don't need to take the `allowNA` and `keepNA` arguments of `nchar()` into account.

crates/jarl-core/src/lints/base/nzchar/nzchar.rs
```diff
@@ -107,3 +115,133 @@ pub fn nzchar(ast: &RBinaryExpression) -> anyhow::Result<Option<Diagnostic>> {
...
+                RSyntaxKind::GREATER_THAN_OR_EQUAL_TO => (
+                    "`nchar(x) >= 0` is always true.",
+                    "Maybe you want `nzchar(x)` instead.",
+                    None,
+                ),
+                RSyntaxKind::LESS_THAN => (
+                    "`nchar(x) < 0` is always false.",
+                    "Maybe you want `!nzchar(x)` instead.",
+                    None,
+                ),
```
etiennebacher, line 198: This is more a "correctness" lint. It would be a bit too much to extract this into another rule for now, but it could be interesting to list functions that can only return positive numeric values and detect this kind of comparison (although this is probably more in scope for a type checker).

etiennebacher (commented):
Thanks! Couple of comments below, one that could be addressed here or in another PR, and one that is not really actionable.

crates/jarl-core/src/lints/base/nzchar/nzchar.rs
Yousa-Mirage: I could add `keepNA = TRUE` to all generated `nzchar()`. However, that is not enough to make the fix for this rule "safe." This is because `x == ""` and `nchar(x) == 0` preserve vector attributes (such as names and matrix information), whereas `nzchar(x)` produces a new vector without attributes. This could lead to issues in certain use cases.

Yousa-Mirage:
OK, added `keepNA = TRUE`.

crates/jarl-core/src/lints/base/nzchar/nzchar.rs
etiennebacher: Ah right, nice catch.

For reference and anyone visiting this later:

``` r
x <- c("asfef", "qwerty", "yuiop[", "b", "stuff.blah.yech")
names(x) <- x
nchar(x) > 0
#>           asfef          qwerty          yuiop[               b stuff.blah.yech 
#>            TRUE            TRUE            TRUE            TRUE            TRUE
nzchar(x)
#> [1] TRUE TRUE TRUE TRUE TRUE
```

etiennebacher (approved):
Thank you!

FYI, there is no need for you to merge main multiple times (unless there are conflicts to fix or important changes on main). Slightly lagging behind main usually doesn't affect the reviews, and merge commits generate notifications that are not useful to me (contrary to actual commits).

PR #706, opened 2026-09-10
fix: Support non-ASCII symbols in package analysis

Yousa-Mirage:
The issue requires a minimal R package to reproduce because `unused_function` compares function definitions and symbol references across package files:

```text
test-jarl/
├── DESCRIPTION
├── NAMESPACE
└── R/
    ├── definitions.R
    └── uses.R
```

`DESCRIPTION`:

```text
Package: demo
Version: 0.0.1
```

`NAMESPACE`:

```r
export(use_functions)
```

`R/definitions.R`:

```r
中文函数 <- function(x) x
未使用 <- function() 1
unused_helper <- function() 2
```

`R/uses.R`:

```r
use_functions <- function(x) 中文函数(x)
```

Run the command in `test-jarl` dir:

```bash
jarl check .
```

Before the fix, `中文函数` is incorrectly reported as defined but never called, alongside the genuinely unused functions. After the fix, only `未使用` and `unused_helper` are reported.

This is because `scan_symbols` scanned raw bytes and recognized only ASCII identifier bytes, while package definition scanning recognized Unicode identifiers, so calls to non-ASCII functions were omitted from occurrence counts and reported as unused.

I use `char_indices()` to preserve UTF-8 byte offsets while scanning Unicode letters and alphanumeric characters, keeping `.` and `_` as valid identifier characters.

---

**Edit:**

The non-ASCII issue with `unused_object` will only occur in Rmd/Qmd:

````qmd
---
title: "test non-ASCII symbols in Rmd/Qmd"
format: html
---

```{r}
变量 <- 1
```

Value is `r 变量` here.
````

Run `jarl check test.qmd --select unused_object`:

```
warning: unused_object
 --> test.qmd:7:1
  |
7 | 变量 <- 1
  | ---- Object `变量` is defined but never used.
  |

Found 1 error.
```

This PR fixed this together.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -27,7 +27,8 @@
 
 ### Bug fixes
 
-* Prevent `unused_function` and `unused_object` from falsely reporting non-ASCII names.
+* Prevent `unused_function` and `unused_object` from falsely reporting non-ASCII
+  names (#706, @Yousa-Mirage).
 
 * Prevent the `nzchar` rule from treating quote characters as empty strings and
   support empty raw string literals (#696, @Yousa-Mirage).
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -27,6 +27,8 @@
 
 ### Bug fixes
 
+* Prevent `unused_function` and `unused_object` from falsely reporting non-ASCII names.
+
 * Prevent the `nzchar` rule from treating quote characters as empty strings and
   support empty raw string literals (#696, @Yousa-Mirage).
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: docs/changelog.md [hunk 1]
```diff
@@ -27,6 +27,9 @@
 
 ### Bug fixes
 
+* Prevent `unused_function` and `unused_object` from falsely reporting non-ASCII
+  names (#706, @Yousa-Mirage).
+
```
etiennebacher, line 32: Is `unused_object` also concerned by this? It relies on `jarl-semantic` so it should be more robust to that.

Yousa-Mirage: Yes, but it will only appear in Rmd/Qmd. I added a reproduction above. You can take it. (AI found this)

PR #708, opened 2026-09-10
fix: Correct Rmd/Qmd suppression quickfix positions

Yousa-Mirage:
Another issue with Rmd/Qmd.

An example `test.qmd`:

````qmd
---
title: "Suppress with jarl-ignore comment"
format: html
---

```{r}
x <- 1
any(is.na(x))
```
````

Click this in IDE:

图片

Then we will get:

````qmd
---
# jarl-ignore any_is_na: <reason>   # The comment is wrongly inserted here.
title: "Suppress with jarl-ignore comment"
format: html
---

```{r}
x <- 1
any(is.na(x))
```
````

The fix is only one line:

```rust
insert_point.line = count_lines_to(file_content, insert_point.offset);
```

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -27,6 +27,8 @@
 
 ### Bug fixes
 
+* Fix LSP `jarl-ignore` quickfix insertion positions in Rmd and Qmd files.
+
 * Prevent the `nzchar` rule from treating quote characters as empty strings and
   support empty raw string literals (#696, @Yousa-Mirage).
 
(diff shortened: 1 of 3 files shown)


docs/changelog.md
```diff
@@ -27,6 +27,9 @@
 
 ### Bug fixes
 
+* Fix LSP `jarl-ignore` quickfix insertion positions in Rmd and Qmd files
+  (#708, @Yousa-Mirage).
```
etiennebacher: Can you merge this with the other item about "language server suppression quickfix positions" below?

Yousa-Mirage: Sure, done.

PR #710, opened 2026-09-12
refactor: Update documentation build

### LINKED ISSUE #703: Refactor Quarto documentation build
In #691 an issue occurred where a line of the `docs/_quarto.yml` file was accidentally erased. Seemingly due to a mix of a VS Code extension and issues with the indentation of the YAML.

When doing the second set of amends for #691 I worked in Positron instead of VS Code to try avoid any the issue reoccurring.

Ahead of the amends, I merged updates from main into the branch, including d1759a4 which updated the YAML including the indentation.

I happened to have `_quarto.yml` open when running `just document` and witnessed the indentation changing in front of me. This led me to the `justfile` and ultimately `docs/make_docs.R`.

I noticed that `make_docs.R` uses the `{yaml}` package to read and write the quarto file. I've previously had issues with the `{yaml}` package not writing out YAML correctly, so thought this might be the culprit. I did some separate tests both with the `_quarto.yml` file and general checks.

```r
cat(yaml::as.yaml(list(
  hello = "world",
  foo = list("bar", "foobar"),
  foo2 = c("bar", "foobar"),
  foo3 = list(bar = c("foo", "foobar")))

#> R OUTPUT           # YAML 1.2 SPEC
#> hello: world       # hello: world
#> foo:               # foo:
#> - bar              #   - bar
#> - foobar           #   - foobar
#> foo2:              # foo2:
#> - foo              #   - foo
#> - bar              #   - bar
#> - foobar           #   - foobar
#> foo3:              # foo3:
#>   foo:             #   foo:
#>   - bar            #     - bar
#>   - foobar         #     - foobar
```

Looking at the repo for `{yaml}`, this seems to be due to an issue with the underlying libYAML C library that it relies on, see r-lib/yaml#13 and r-lib/yaml#102.

You can use the argument `indent.mapping.sequence = TRUE` within the `write_yaml()` call to implement the local fudge that the package has implemented. But I'd probably suggest switching to [`{yaml12}`](https://posit-dev.github.io/r-yaml12/).

I also noticed that line 88 inserts/replaces the filters, even though these are already included in the file that's read in, so this feels like a redundant line.

https://github.com/etiennebacher/jarl/blob/3cf73283921dcee977b7234d887cdf0c208190f4/docs/make_docs.R#L88

Happy to make a PR if you'd like - if so do you have a view on `{yaml}` or `{yaml12}`

mattkerlogue:
Refactoring of the process for building of Jarl documentation, closes #703. In summary:

- Switched to using `{yaml12}` for reading Quarto config.
- Documentation blocks in `<rulename>.rs` must start with `<!-- docs: start -->` and end with `<!-- docs: end -->`
- Documentation for configuration moved to the `options.rs` file associated with a given rule and included in the rule's documentation not the general configuration file documentation.
- Added pre-flight check for required packages to the documentation build process.

See #703 for full discussion/details.


crates/jarl-core/src/lints/base/assignment/assignment.rs
```diff
@@ -15,11 +16,20 @@ use biome_rowan::AstNode;
 /// use `=` without problems. This rule only ensures the consistency of the
 /// assignment operator in a project.
 ///
-/// Set the following option in `jarl.toml` to use `=` as the preferred operator:
-///
+/// The `lint.argument.operator` option in `jarl.toml` must be set for this
```
etiennebacher, line 19: Shouldn't these docs be stored in `assignment/options.rs`?

crates/jarl-core/src/lints/base/implicit_assignment/options.rs
```diff
@@ -23,12 +23,31 @@ const DEFAULT_SKIPPED_FUNCTIONS: &[&str] = &[
...
+/// Function names in `skipped-functions` or `extend-skipped-functions` also match
+/// namespaced calls, e.g. `skipped-functions = ["list2"]` will ignore `list2()` and
+/// `rlang::list2()`.
+/// 
+/// Default: `skipped-functions = ["alist", "expect_error", "expect_warning", "expect_message",
+/// "expect_silent", "expect_defunct", "expect_deprecated", "expect_snapshot",
+/// "expect_no_condition", "expect_no_warning", "expect_no_error", "expect_no_message",
+/// "quote", "suppressMessages", "suppressWarnings", "try"]`
+/// (`expect_`functions come from the `testthat` package, except `expect_defunct` and
+/// `expect_deprecated` which come from the `lifecycle` package)
```
etiennebacher, line 40: I know this is formatted like this in the existing docs but can we put one function per line, as in `missing_argument/options.rs`?

crates/jarl-core/src/lints/base/unreachable_code/options.rs
```diff
@@ -6,12 +6,27 @@ use crate::rule_options::resolve_with_extend;
...
+/// Use `stopping-functions` to fully replace the default list of functions that are
+/// considered to stop execution (never return). Use `extend-stopping-functions` to
+/// add to the default list. Specifying both is an error.
+/// 
+/// Function names in `stopping-functions` or `extend-stopping-functions` also match
+/// namespaced calls, e.g. `stopping-functions = ["abort"]` will consider `abort()`
+/// and `rlang::abort()` as stopping functions.
+/// 
+/// Default: `stopping-functions = ["stop", ".Defunct", "abort", "cli_abort",
+/// "q", "quit"]`.
```
etiennebacher, line 19: Same as above, can this expanded with one function per line?

crates/jarl-core/src/lints/base/unused_object/options.rs
```diff
@@ -2,12 +2,33 @@ use std::collections::HashSet;
...
+/// namespaced calls, e.g. `skipped-functions = ["expect_error"]` will allow
+/// `expect_error()` and `testthat::expect_error()`.
+/// 
+/// Only the direct argument position counts: an assignment nested in a block or in
+/// a function defined inside the call is an ordinary local and is still reported.
+/// 
+/// Default: `skipped-functions = ["expect_error", "expect_warning",
+/// "expect_message", "expect_silent", "expect_defunct", "expect_deprecated",
+/// "expect_snapshot", "expect_no_condition", "expect_no_warning",
+/// "expect_no_error", "expect_no_message"]`
```
etiennebacher, line 21: Same

crates/jarl/tests/integration/rule.rs
```diff
@@ -109,11 +109,20 @@ fn test_known_rule_disabled_by_default() -> anyhow::Result<()> {
     use `=` without problems. This rule only ensures the consistency of the
     assignment operator in a project.
 
-    Set the following option in `jarl.toml` to use `=` as the preferred operator:
+    The `lint.argument.operator` option in `jarl.toml` must be set for this
```
etiennebacher, line 112: Shouldn't this be in a "Configuration options" section?

docs/rules/duplicated_arguments.md
```diff
@@ -21,3 +22,28 @@ should typically be avoided.
 ```r
 list(x = 1, x = 2)
 ```
+
+---
+
+## Configuration options
```
etiennebacher, line 28: I would rather have these options appear earlier in the rule docs, so that the structure for the rule docs follows:

- what it does
- why is this bad?
- ... possibly other sections ...
- config options
- examples
- references

Note that this should probably use the `## Examples` marker to know where to insert the config options instead of using `## Why is this bad?` because there can be extra sections after `## Why is this bad?` (e.g. in `unused_object`).

docs/contributing-rule-example.md
```diff
@@ -538,6 +540,103 @@ Looking at tests for `list2df`, there are four blocks:
...
+/// <!-- docs: end -->
+impl Violation for List2Df {
+    ...
+}
+```
+
+A documentation block must declare the version added (in `MAJOR.MINOR.PATCH` format).
+It should also have a "What it does" section that describes what the rule does, most rules also have a "Why is this bad" section to explain the rationale for the rule.
+Be helpful to end user and include an "Example" that illustrates bad code that will be flagged by the rule and good code that will not be flagged by the rule.
+You may also wish to include references, for example to help commands within R or external websites. Section headers should use level 2 heading markers (`##`).
```
etiennebacher, line 607:
```suggestion
It should also have the following subsections:

 - "What it does" to explain what the rule does
 - "Why is this bad" to explain the rationale for the rule and potential caveats
 - "Example" to illustrate code that will be flagged by the rule and code that will not be flagged by the rule
 - Potentially a "References" section, such as `?<function>` if more details about the `<function>` concerned by the rule can be found in R docs, or links to external websites. 
```

docs/contributing-rule-example.md
```diff
@@ -538,6 +540,103 @@ Looking at tests for `list2df`, there are four blocks:
...
+
+::: {.callout-note title = "Click to see how to add documentation for TOML options" collapse = true}
+The documentation for any TOML options should be included in the `options.rs` file.
+Like the documentation for the rule itself, the documentation should be in a block that is commented out with three slashes (`///`) and uses the `<!-- docs: start -->` and `<!-- docs: end -->` markers.
+
+The options documentation should explain the option(s) that can be set in the configuration file and their possible values.
+It should note the effect(s) of choosing different options and declare any defaults or conflicts.
+The documentation should also include an example of how to set the preferred option in the `jarl.toml` configuration file.
+
+The options documentation will be included after the main documentation for the rule under a section called "Configuration options".
```
etiennebacher, line 617:
```suggestion
The options documentation will be included in the rule documentation under a section called "Configuration options".
```

docs/contributing-rule-example.md
```diff
@@ -538,6 +540,103 @@ Looking at tests for `list2df`, there are four blocks:
 Since we have snapshot tests, we first need to run `cargo insta test` to generate the snapshots and then `cargo insta review` to review and validate them.
 After that, run `cargo test` to ensure that all tests pass.
 
+### Add documentation
```
etiennebacher, line 543: I think this should live only in `contributing.md` instead of being duplicated here as well. Just having a single link referring to the section in `contributing.md` should be enough.

etiennebacher (changes requested):
Thank you, this looks great and I think this will improve a lot both the current documentation and the documenting process for contributors. I have a few comments and questions to address below.

mattkerlogue:
Thanks, those comments all seem sensible, will progress them.

Tbh, I didn't think just appending the config options at the end of the document felt right but wasn't sure where in the flow of the document you'd like them to appear. Agree inserting before Examples/References feels like the right place.

This will also solve the issue with the `assignment` rule, as far as I can tell it's the only rule that explicitly requires a config option to be set for it to operate. The existing rule docs make reference to the config option setting so felt weird for these to be duplicated. But in moving config options further up the document flow then can re-write and split the documentation so it follows the new pattern.

mattkerlogue:
I've just pushed two commits that implement the reflow and update the contributing guidance.

I noticed when going through the documentation that there's inconsistencies in the formatting of fenced code blocks. There's a mix of fenced code blocks being surrounded by blank lines and sometimes where they're not.

https://github.com/etiennebacher/jarl/blob/1ce18ec619be6275a4425a05f69a29c31581e1ea/crates/jarl-core/src/lints/base/if_not_else/if_not_else.rs#L38-L40

https://github.com/etiennebacher/jarl/blob/1ce18ec619be6275a4425a05f69a29c31581e1ea/crates/jarl-core/src/lints/base/grepv/grepv.rs#L40-L41

The Commonmark standard says blank lines are not required but it appears that sometimes there are issues with Quarto/Air when code blocks don't have blank lines surrounding them see [this issue](https://github.com/quarto-dev/quarto-cli/issues/12255) in the Quarto repo. The markdownlint extension for VSCode/Positron also flags this is an issue in the markdown files.

While I've not actually encountered any issues from this discrepancy for clarity it's probably worth being consistent through the docs.

Another inconsistency I noticed is that there seems to be docs that use asterisks and others that use hyphens for unordered lists, but this is purely stylistic since both are permitted in the markdown standards, and it'd be a bit of pain conforming all of those to a single style. So don't propose doing anything about that.

crates/jarl-core/src/lints/base/assignment/options.rs
```diff
@@ -1,9 +1,27 @@
...
-///
-/// Use `operator` to specify which assignment operator to enforce.
-/// Valid values are `"<-"` (the default) and `"="`.
+/// <!-- docs: start -->
+/// The `lint.assignment.operator` option in `jarl.toml` must be set for this
+/// rule to be checked. It takes a single value (`"<-"` or `"="`) indicating
+/// the preferred assignment operator in the files to check.
+/// If `operator = "<-"` then any use of the `"="` operator to assign values
+/// will be reported, and vice-versa. To stop checking this rule remove this
+/// setting from the configuration file.
```
etiennebacher, line 9:
```suggestion
/// will be reported, and vice-versa.
```

The "universal" way to not check a rule is to remove it from selected rules or to add it to the excluded rules. I'd rather not add this kind of special case.

You'll need the update some snapshots following this.

crates/jarl-core/src/lints/base/quotes/quotes.rs
```diff
@@ -51,13 +52,9 @@ impl ParsedString<'_> {
...
-/// quotes. Therefore, by default, this rule expects double quotes (`"`).
-///
-/// To prefer single quotes, set this in `jarl.toml`:
-/// ```toml
-/// [lint.quotes]
-/// quote = "single"
-/// ```
+/// quotes. Therefore, by default, this rule expects double quotes (`"`). See
+/// configuration options (below) for details about how to configure Jarl to
+/// prefer single quotes.
```
etiennebacher:
```suggestion
/// quotes. Therefore, by default, this rule expects double quotes (`"`). 
```

docs/reference/_config-file.md
etiennebacher, line 1: This needs to be updated since Jarl now also accepts `.jarl.toml`, you just need to apply these changes: https://github.com/etiennebacher/jarl/commit/88723fa12bd2b1dbdd225c437a2805e853c01d24#diff-0a1813df370063163f6ee3e5ca08fbe8ecace85b652f69aa8ed2b4e47aa36b8e

docs/reference/config-file.md
```diff
@@ -285,248 +290,19 @@ fix-roxygen = false
...
-...
-
-[lint.unused_object]
-# Also allow an unused assignment passed straight to `my_expect()`.
-extend-skipped-functions = ["my_expect"]
-```
+The operation of some rules can be customised with arguments in the
+configuration file, see the relevant rule documentation page for details:
+
+* [`assignment`](../rules/assignment.md#configuration-options)
```
etiennebacher, line 296: Very cool to have this list!

docs/contributing.md
```diff
@@ -65,6 +65,45 @@ Adding a new rule requires four main steps:
...
 
+### Documenting a new rule
+
+The documentation used in `jarl rule <rulename>` and on the Jarl website is generated from the documentation block in the `<rulename.rs>` file.
+
+The documentation block is positioned after any imports and declarations but before any of the code that operates the rule checks.
+The documentation block must be commented out using three slashes (`///`), it must start with the marker `<!-- docs: start -->` and end with the marker `<!-- docs: end -->`.
+Commented content outside of these markers will be ignored.
+The content of the documentation block uses standard markdown notation.
+Although the documentation website is built using Quarto do not include any executable code as the documentation is extracted into plain markdown files.
```
etiennebacher, line 76:
```suggestion
Although the documentation website is built using Quarto, do not include any executable code as the documentation is extracted into plain markdown files.
```

docs/contributing.md
```diff
@@ -65,6 +65,45 @@ Adding a new rule requires four main steps:
...
+///
+/// [...]
+/// <!-- docs: end -->
+```
+
+The first line of a documentation block must declare the version added (in `MAJOR.MINOR.PATCH` format).
+The documentation block should have the following subsections (each using a level 2 (`##`) markdown header):
+
+* "What is does" to explain what the rule does
+* "Why is this bad?" to explain the rational for the rule and potential caveats
```
etiennebacher, line 94:
```suggestion
* "Why is this bad?" to explain the rationale for the rule and potential caveats
```

docs/contributing.md
```diff
@@ -65,6 +65,45 @@ Adding a new rule requires four main steps:
...
+* "Example" to illustrate code that will be flagged by the rule and code that will not be flagged by the rule
+* Potentially a "References" section, such as `?<function>` if more details about the `<function>` the rule relates to can be found in R help, and/or links to external websites.
+
+If the rule has configuration options, these should be documented in the rule's `options.rs` file and will be included in the rule's documentation page under the subsection heading "Configuration options".
```
etiennebacher, line 98:
```suggestion
If the rule has configuration options, these should be documented in the rule's `options.rs` file and will be automatically included in the rule's documentation page under the subsection heading "Configuration options".
```

docs/contributing.md
```diff
@@ -65,6 +65,45 @@ Adding a new rule requires four main steps:
...
+Options documentation should also be commented out using three slashes and include `<!-- docs: start -->` and `<!-- docs: end -->` markers.
+The options documentation should explain the purpose of any option(s), their possible values and their effect(s), it should also have the following level 3 (`###`) subsections:
```
etiennebacher, line 100:
```suggestion
The options documentation should explain the purpose of any option(s), their possible values and their effect(s).
It should also have the following level 3 (`###`) subsections:
```

docs/make_docs.R
```diff
@@ -1,95 +1,53 @@
...
-      call. = FALSE
-    )
-  }
+### get list of lint rules ----
 
-  start <- grep("## What it does", content, fixed = TRUE)
-  end <- grep("^(impl Violation for|fn |pub fn|// )", content) - 1
-  end <- end[end > start]
-  end <- end[1] # could be several "pub fn"
+lints_dir <- "crates/jarl-core/src/lints" #
```
etiennebacher, line 11:
```suggestion
lints_dir <- "crates/jarl-core/src/lints"
```

docs/make_docs.R
```diff
@@ -1,95 +1,53 @@
...
+  contents = list(
+    "rules.qmd",
+    list(
+      section = "List of rules",
+      contents = paste0("rules/", sort(rule_docs))
+    )
   )
+))
+
+# use format_yaml until https://github.com/posit.dev/r-yaml12/issues#40 fixed
```
etiennebacher, line 49:
```suggestion
# use format_yaml until https://github.com/posit-dev/r-yaml12/issues/40 is fixed
```

docs/make_docs_helpers.R
```diff
@@ -0,0 +1,170 @@
...
+
+  gsub(
+    "^///\\s?",
+    "",
+    rust_script[(doc_markers$start + 1):(doc_markers$end - 1)]
+  )
+}
+
+create_rule_md <- function(rule, lints_dir) {
+  # rule meta
```
etiennebacher, line 26:
```suggestion
create_rule_md <- function(rule, lints_dir) {
```

etiennebacher (changes requested):
Thank you, this is great! 

Some minor things to fix below and after that it should be good to go! Also please fix the clippy failure.

> The Commonmark standard says blank lines are not required but it appears that sometimes there are issues with Quarto/Air when code blocks don't have blank lines surrounding them see https://github.com/quarto-dev/quarto-cli/issues/12255 in the Quarto repo. The markdownlint extension for VSCode/Positron also flags this is an issue in the markdown files.

This doesn't seem to cause any issue since we produce md files and not qmd ones, so feel free to harmonize this but I wouldn't consider this a blocker for this PR.

> Another inconsistency I noticed is that there seems to be docs that use asterisks and others that use hyphens for unordered lists, but this is purely stylistic since both are permitted in the markdown standards, and it'd be a bit of pain conforming all of those to a single style. So don't propose doing anything about that.

Agreed, this has no consequences.

mattkerlogue:
Just a quick note to say thanks for the comments and that I'm having a rather busy couple of weeks, but I should be able to get back to finishing this off next Monday.

etiennebacher:
Thanks for the update, maybe I'll merge this as-is and make a follow up PR to address my (minor) comments.

mattkerlogue:
Up to you. I can probably squeeze in sorting it out tomorrow am tbh. But know that I've definitely got time on Monday I can devote to it.

etiennebacher:
@mattkerlogue I don't know why I couldn't push directly to your branch so I fixed the conflicts and addressed my comments above in #730, and I now close this.  Many thanks again, this is great!

PR #712, opened 2026-09-12
fix: Preserve linting for roxygen examples with nested braces

Yousa-Mirage:
A `test.R` like this:

```r
#' @examples
#' \dontrun{
#' identity({
#'   any(is.na(x))
#' }
#' )
#' }

foo <- 5
```

Then:

```bash
$jarl check test.R

All checks passed!
```

The issues in `@examples` are ignored and no lints.

This is because the old logic matched every standalone `}` as a roxygen wrapper close without tracking ordinary braces or persistent string state. For example:

```r
\dontrun{
if (x) {
  any(is.na(x))
}  # With the old logic this line will be deleted.
}  # But this line is the same, so the chunk is still correct.
```

But for this code:

```r
\dontrun{
identity({
  any(is.na(x))
}  # This is deleted wrongly.
)
}  # This is left.
```

It will become this:

```r
identity({
  any(is.na(x))
)
}
```

The code in chunk is broken and then failed to parse.

I added a shared state machine that tracks code and wrapper braces across lines while ignoring strings, raw strings, and comments.

### DIFF
--- a/crates/jarl-core/src/roxygen.rs
+++ b/crates/jarl-core/src/roxygen.rs
@@ -207,6 +207,8 @@ fn roxygen_macro_lines_to_remove<T: AsRef<str>>(code_lines: &[T]) -> Vec<usize> 
                 match mode {
                     StringMode::Quoted(quote) => {
                         if bytes[pos] == b'\\' {
+                            // A backslash escapes the next byte, including a
+                            // line-ending escape when it is the last byte.
                             pos += 1;
                             if pos < bytes.len() {
                                 pos += 1;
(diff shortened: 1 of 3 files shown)


etiennebacher:
Thank you, I'd rather put this on hold for now since there is work being done in Air to support `@examples`. Ideally I'd like to remove the custom infrastructure in Jarl to use theirs, although it is not sure they will support `\dontrun` and friends, in which case this PR will be useful. 

Given that there are no ecosystem violations, I don't think this is an issue that needs to be solved now.

PR #726, opened 2026-09-19
feat: Custom messages for undesirable functions (#655)

### LINKED ISSUE #655: Feature: Allow `undesirable_function` to take messages for `extend-functions`
I'd like to be able to use the `undesirable_function` linter to ban additional functions via `extend-functions` but also indicate what should be used in its place or provide other information about why it is banned.

For example, I would like to be able to do the following:

```toml
extend-functions = [
  { setwd = 'Use here::here().' },
  { sprintf = 'Use glue::glue().' },
  { transmute = 'Use mutate(.keep = "none").' },
]
```

where `setwd` would indicate a different (negative) type of workflow being introduced, `sprintf` is harder to read than `glue` if I'm already relying on it, and `transmute` seems to be GPT-5.6's favorite `dplyr` function because it has no sense of superseding. Of course, a user might want to extend this with other things that fit their workflow depending on the project, like listing error and warning functions with `cli`.

lintr allows similar messages:

```
lint(
  text = "log10(x)",
  linters = undesirable_function_linter(fun = c("log10" = "use log()"))
)
```

I've built a [proof of concept](https://github.com/christopherkenny/jarl/tree/undesirable-function-messages) that demos this and I'm happy to make a PR if you think it is useful.

christopherkenny:
Provides support to make comments of the form:

```toml
extend-functions = [
  { setwd = 'Use here::here().' },
  { sprintf = 'Use glue::glue().' },
  { transmute = 'Use mutate(.keep = "none").' },
]
```

Comments remain optional and the custom comments are appended, rather than replacing the base message.

Closes #655.

### DIFF
--- a/crates/jarl-core/src/lints/base/undesirable_function/mod.rs
+++ b/crates/jarl-core/src/lints/base/undesirable_function/mod.rs
@@ -4,6 +4,7 @@ pub(crate) mod undesirable_function;
 #[cfg(test)]
 mod tests {
     use crate::lints::base::undesirable_function::options::ResolvedUndesirableFunctionOptions;
+    use crate::lints::base::undesirable_function::options::UndesirableFunctionEntry;
     use crate::lints::base::undesirable_function::options::UndesirableFunctionOptions;
     use crate::rule_options::ResolvedRuleOptions;
     use crate::settings::{LinterSettings, Settings};
@@ -76,7 +77,7 @@ mod tests {
     #[test]
     fn test_custom_functions() {
         let settings = settings_with_options(UndesirableFunctionOptions {
-            functions: Some(vec!["debug".to_string()]),
+            functions: Some(vec![UndesirableFunctionEntry::Name("debug".to_string())]),
             extend_functions: None,
         });
 
@@ -102,7 +103,7 @@ mod tests {
     fn test_extend_functions() {
         let settings = settings_with_options(UndesirableFunctionOptions {
             functions: None,
-            extend_functions: Some(vec!["debug".to_string()]),
+            extend_functions: Some(vec![UndesirableFunctionEntry::Name("debug".to_string())]),
         });
 
         // "browser" is still in the defaults -> lints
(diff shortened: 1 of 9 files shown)


christopherkenny:
Two small bits of context worth mentioning:

- It may make sense to add default messages to some undesirable functions
- This adds the `pkg::fn()` instead of just `fn()` to the printed message. Sol suggested it and I thought it looked nice, so it's in there. Can be dropped if you disagree

docs/reference/config-file.md
```diff
@@ -450,6 +450,28 @@ Default: `skipped-functions = []`
...
 ```
 
+### `undesirable_function`
+
+Use `functions` to replace the default list of undesirable functions, or
+`extend-functions` to add to it. Specifying both is an error. The default list
+contains `browser`.
+
+Entries can be function-name strings or inline tables mapping function names to
+custom suggestion text. A mapped function is included in the selected list,
```
etiennebacher, line 460: Can you add an example where `extend-functions` is just a list of of strings?

docs/reference/config-file.md
```diff
@@ -450,6 +450,28 @@ Default: `skipped-functions = []`
...
+```toml
+[lint]
+...
+
+[lint.undesirable_function]
+extend-functions = [
+  { setwd = 'Use here::here().' },
+  "sprintf",
+  { transmute = 'Use mutate(.keep = "none").' },
+]
```
etiennebacher, line 472:
```suggestion
[lint.undesirable_function]
extend-functions = [
  { setwd = 'Use `here::here()`.' },
  "sprintf", # Providing a suggestion is not mandatory.
  { transmute = 'Use `mutate(.keep = "none")`.' },
]
```

docs/changelog.md
```diff
@@ -166,6 +166,9 @@
   to list functions whose arguments are allowed to contain the `T` and `F`
   symbols (#542).
 
+* `undesirable_function` now accepts inline tables in `functions` and
+  `extend-functions` to attach custom suggestions to undesirable functions.
```
etiennebacher:
```suggestion
* `undesirable_function` now accepts inline tables in `functions` and
  `extend-functions` to attach custom suggestions to undesirable functions (#726).
```

docs/reference/config-file.md
```diff
@@ -450,6 +450,28 @@ Default: `skipped-functions = []`
 skipped-functions = ["foo"]
 ```
 
+### `undesirable_function`
+
+Use `functions` to replace the default list of undesirable functions, or
+`extend-functions` to add to it. Specifying both is an error. The default list
+contains `browser`.
```
etiennebacher, line 457: There is no default list of undesirable functions to report, so this should only take a `functions` option. (`unused_function` is another rule like that.)

etiennebacher (changes requested):
Thanks, this looks good!

The main comment below is that the TOML section should only take a `functions` option and not `extend-functions`.

christopherkenny:
Starting to move on this, but can you clarify one thing: wouldn't browser() be a default undesirable function?

etiennebacher:
Ah yes indeed, you can dismiss my comment above about that.

crates/jarl/tests/integration/toml_rule_args.rs
```diff
@@ -768,6 +768,87 @@ quote = "foo"
     Ok(())
 }
 
+// undesirable_function ----------------------------------------
```
etiennebacher, line 771: I would like to see more validation of this new syntax, for instance the following cases should error with a good message:

```
extend-functions = [
  { 1 = 'Use here::here().' }
]
extend-functions = [
  { true = 'Use here::here().' }
]
extend-functions = [
  { setwd = 1 }
]
extend-functions = [
  { setwd = true }
]
// forbid empty names
extend-functions = [
  { "" = true }
]
// probably forbid whitespace?
extend-functions = [
  { "  setwd  " = true }
]
```

crates/jarl-core/src/lints/base/undesirable_function/mod.rs
```diff
@@ -133,4 +134,48 @@ mod tests {
         "
         );
     }
+
+    #[test]
+    fn test_custom_messages() {
+        let options: UndesirableFunctionOptions = toml::from_str(
+            r#"
+            extend-functions = [
+                { setwd = 'Use here::here().' },
```
etiennebacher, line 143: We should also that test `{ "base::setwd" = 'Use here::here().' }` works and that it only reports `base::setwd()` calls and not `setwd()` (this should also be documented).

etiennebacher (changes requested):

